PR #72, opened 2024-03-25
update tmux commands

wakamex:
on the latest tmux (3.4 and 3.5-next) split-window without `-l` fails unless you have a % after the number.
https://github.com/tmux/tmux/issues/3836#issuecomment-1942216630

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,14 +3,11 @@ dev:
 	@docker compose up -d redis
 	@docker compose up -d postgres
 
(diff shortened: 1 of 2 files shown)


murderteeth:
Have one problem, I get this tmux layout when I run the new make\setup_devenv,
no

but expecting this,
yes

Is it the same for you?

wakamex:
yup, I thought you just liked a weird layout. fixed it
image

PR #213, opened 2025-01-31
misc(bun): migrate to use `bun`

matheus1lva:
# What was done
1. Migrate to use bun as package manager
2. leverage speed of bun test runner

Edit:
Due to some problems on:
1. bun test runner and it not respecting serial/parallel config
2. bun not working correctly with `ellide-lines` correctly, neither on cli nor on their toml file (https://github.com/oven-sh/bun/issues/17918) we have decided to go back to yarn for command running and leverage bun for installation for now!

Edit 2:
1. Added testcontainer so everybody has the same env in terms of database and redis. Allowing even to run on ci!


murderteeth (changes requested):
heyo, thanks for the pr!

- i just configured eslint in main, please rebase and update your code to match the style. i added scripts to the root package.json for this. should be sth like `bun lint:fix:ingest` 

https://github.com/yearn/kong/pull/215
https://github.com/yearn/kong/pull/216

- make sure that `make dev` also uses bun instead of yarn. also make sure you can run `make dev` =)

Makefile
```diff
@@ -10,22 +10,19 @@ dev:
 	./setup_devenv.sh
 
 	# Run commands
-	@tmux send-keys -t kongdevenv:0.0 'PORT=$${PORT:-3001} yarn workspace web dev' C-m
-	@tmux send-keys -t kongdevenv:0.1 'yarn workspace ingest dev' C-m
-	@tmux send-keys -t kongdevenv:0.2 'yarn workspace terminal dev' C-m
-	@tmux send-keys -t kongdevenv:0.3 'sleep 6 && yarn workspace db migrate up && PGPASSWORD=password psql --host=localhost --port=5432 --username=user --dbname=user' C-m
+	@tmux send-keys -t kongdevenv:0.0 'PORT=$${PORT:-3001} bun --filter web dev' C-m
```
murderteeth: re: Makefile
using `bun --filter <package> <command>` isn't working right for me. try this instead:

```

-       @tmux send-keys -t kongdevenv:0.0 'PORT=$${PORT:-3001} bun --filter web dev' C-m
-       @tmux send-keys -t kongdevenv:0.1 'bun --filter ingest dev' C-m
-       @tmux send-keys -t kongdevenv:0.2 'bun --filter terminal dev' C-m
-       @tmux send-keys -t kongdevenv:0.3 'sleep 6 && bun --filter db migrate up && PGPASSWORD=password psql --host=localhost --port=5432 --username=user --dbname=user' C-m
+       @tmux send-keys -t kongdevenv:0.0 '(cd packages/web && PORT=$${PORT:-3001} bun dev)' C-m
+       @tmux send-keys -t kongdevenv:0.1 '(cd packages/ingest && bun dev)' C-m
+       @tmux send-keys -t kongdevenv:0.2 '(cd packages/terminal && bun dev)' C-m
+       @tmux send-keys -t kongdevenv:0.3 'sleep 6 && (cd packages/db && bun migrate up) && PGPASSWORD=password psql --host=localhost --port=5432 --username=user --dbname=user' C-m

```

when you run `make dev` it should look like this:
image

but when i run make dev with bun --filter i get this:
image

test/setup.ts
murderteeth, line 1: do we need this setup file?

package.json
```diff
@@ -10,6 +10,7 @@
   ],
   "scripts": {
     "clean": "rm -rf node_modules packages/**/node_modules packages/**/dist packages/**/.next",
+    "test": "docker-compose up --force-recreate -d & wait-on tcp:localhost:5432 && wait-on tcp:localhost:6379 && sleep 8 && yarn workspace db migrate up && bun test --preload ./test/setup.ts --timeout 1500000",
```
murderteeth: still have a yarn reference here, `yarn workspace db migrate up`

murderteeth (changes requested):
what command are you using to test? when i use `bun test` from the project root i get 30 failed tests, various path and connection issues. fwiw i also tested my local config against main using yarn and all tests currently pass. 

image


re price magic (yprice), we have this disabled for now. so my env is just

```

YPRICE_ENABLED=
YPRICE_API=
YPRICE_API_X_SIGNER=
YPRICE_API_X_SIGNATURE=

```

also, update README as well with new bun commands please! kong starts sooo much faster now 😁

matheus1lva:
> what command are you using to test? when i use `bun test` from the project root i get 30 failed tests, various path and connection issues. fwiw i also tested my local config against main using yarn and all tests currently pass.
> 
> image
> 
> re price magic (yprice), we have this disabled for now. so my env is just
> 
> ```
> 
> YPRICE_ENABLED=
> YPRICE_API=
> YPRICE_API_X_SIGNER=
> YPRICE_API_X_SIGNATURE=
> ```
> 
> also, update README as well with new bun commands please! kong starts sooo much faster now 😁

yep, this is what i wanted to check it out, and if having yprice was an issue. Now i can focus on fixing them.

matheus1lva:
CleanShot 2025-02-09 at 18 01 49@2x

Noice its passing now!

Time to update the README file.

matheus1lva:
I had to still leave terminal untouched using `ts-node` because bun is not friendly with processes that have a while, and it simply exits. The only solution is to leave as is.

Although I have xp with infra, I will leave the render config for you, cuz I've never used render.

matheus1lva:
It might need to setup the .env on github secrets for the test workflow to run.

https://streamable.com/1krg7p?src=player-page-share - there's the testing on the test command, after a cleanup.

murderteeth (changes requested):
## make dev

make dev is still broke here. fn i think its find to just use `(cd package/name && bun run dev)`.. bun is still new! it's ok if we cant do everything the "bun way" esp here i think

image


## bun --filter ingest test

tests run and pass! 

but.. the test output seems to be going to this little "box" and not streaming direct to standard.out maybe?.. idk but as a person running these tests, i want to see the full output.  (btw also hangs during test teardown. existing bug tho! we'll fix that another time.. i just ctrl-c to end tests currently.)

image

## bun --filter lib test

test run and pass! great work =)

matheus1lva:
Ah thats fine, yeah we are discovering it together :D

I learn that https://bun.sh/blog/bun-v1.1.43#elide-lines-n-controls-filter-output-line-length is something that disables the lines from showing entirely, I've added that to the make itself, might be work adding to each package as well.

Ill update it and also keep the readme with that BIG caveat from bun :D

matheus1lva:
I've updated readme and all the commands around that used bun run!

matheus1lva:
Reverted Makefile and added a comment on the bottom of the readme

murderteeth (changes requested):
two small things! and we can merge

## README

readme already has a `Dev Notes` section here https://github.com/yearn/kong?tab=readme-ov-file#dev-notes ..pls move your bun comments there

## Makefile

lets use this pls

```


clean:
	@bun clean

dev:
	@docker compose up -d redis
	@docker compose up -d postgres

	# Set up tmux environment
	./setup_devenv.sh

	# Run commands
	@tmux send-keys -t kongdevenv:0.0 '(cd packages/web && PORT=$${PORT:-3001} bun dev)' C-m
	@tmux send-keys -t kongdevenv:0.1 '(cd packages/ingest && bun dev)' C-m
	@tmux send-keys -t kongdevenv:0.2 '(cd packages/terminal && bun dev)' C-m
	@tmux send-keys -t kongdevenv:0.3 'sleep 6 && (cd packages/db && bun migrate up) && PGPASSWORD=password psql --host=localhost --port=5432 --username=user --dbname=user' C-m

	@tmux selectp -t 2
	@tmux attach-session -t kongdevenv

	# This happens when tmux session ends
	@docker compose down

test:
	@bun run --elide-lines 0 --filter lib test
	@bun run --elide-lines 0 --filter ingest test


down:
	@docker compose down
	-@tmux kill-session -t kongdevenv

```

murderteeth (approved):

PR #214, opened 2025-02-02
feat(errors): add errors panel for indexing errors

matheus1lva:
I've added a component on the main dashboard to showcase all the indexing errors. Well, despite the default be a graphql API, next.js was having problems dealing with some bull dependencies over the graphql route, I had to do that over plain HTTP so It could handle it better. 

https://github.com/user-attachments/assets/5014a2db-8b7d-4fef-8314-05ee2a30520a


packages/web/app/api/errors/route.ts
```diff
@@ -0,0 +1,29 @@
+import { Queue } from "bullmq";
+import { bull } from "../mq/bull";
+
+// do something with this later
+export const q = {
```
matheus1lva, line 5: well, we have a decision to be made. We need to keep track of the queues names. Where would this be the best place to be since we are consuming over the `ingest` package and over here?

PR #218, opened 2025-02-09
feat(logs): add log processing and ai summarization

matheus1lva:


matheus1lva:
@murderteeth i've added a test case using bun already, so i don't have to write it using mocha and besides, we can merge the #213 first!

matheus1lva:
@murderteeth i've updated the test for it; added a few more guardrails so we dont run it when its not necessary; and fixed a mocha <> test containers orchestration issue i've found!

PR #219, opened 2025-02-09
fix: null for tvl

### LINKED ISSUE #212: Make sure timeseries data shows nulls instead of zero when data is not available
reported from discord, https://discord.com/channels/734804446353031319/1334168008192430140

Hi people, just to let you know kong is missing some values (which seem to be populated with zeros, would rather have received NaNs and would have detected the issue more easily):

```

query Tvls($chainId: Int!, $address: String, $limit: Int) {
    tvls(chainId: $chainId, address: $address, limit: $limit) {
        time
        priceUsd
    }
}

```

This query returns zeros in priceusd.

You can test with 0x4d3ceBA4349ADB06d2De8EBD2F9320A61303aD81
but it's true for many vaults.
Would propose to keep NaNs in the db, if impossible to have a proper measurement of the price, thank you

matheus1lva:
closes #212 

Draft until i seed my whole db so I can find this address T.T

### DIFF
--- a/packages/web/app/api/gql/resolvers/tvls.ts
+++ b/packages/web/app/api/gql/resolvers/tvls.ts
@@ -47,7 +47,7 @@ const tvls = async (_: object, args: {
       t.period,
       t.block_number,
       t.time,
-      COALESCE(p.price_usd, 0) AS price_usd,
+      p.price_usd,
       COALESCE(p.price_source, 'na') AS price_source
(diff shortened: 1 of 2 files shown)


matheus1lva:
well, despite having the db access, the addresses shown on the issue don't return anything. Despite, the solution works.

@murderteeth ready to review!

PR #220, opened 2025-03-09
fix: null value for correct dataset

matheus1lva:
solve #212

### DIFF
--- a/packages/ingest/abis/erc4626/timeseries/pps/hook.ts
+++ b/packages/ingest/abis/erc4626/timeseries/pps/hook.ts
@@ -35,7 +35,7 @@ export default async function process(chainId: number, address: `0x${string}`, d
   const pps = await _compute(vault, blockNumber)
   if (!pps) return []
 
-  return OutputSchema.array().parse([{
+  return OutputSchema.array().parse([{ 
     chainId, address, label: data.outputLabel, component: 'raw',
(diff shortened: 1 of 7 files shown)


matheus1lva:
cc @murderteeth

murderteeth (changes requested):
we have to think about this a bit more. for example,

### tvl resolver / prices
it's ok to replace 0 with null (but note, its' not ok to replace 0 x null for the value field)

### timerseries resolver 
right now we have these timeseries: APY, TVL, PPS. for PPS it's ok to replace 0 x null. but not ok for APY or TVL.

instead of replacing 0 x null in the resolvers, we may need to make updates at the sources. for example, look at TVL code and make sure it only logs 0 when it should really be 0. 

but! i reviewed the ticket and see the user specifically had an issue with prices on the tvl resolver. so i think we're good actually! lets just delete the 0 x null logic in the timeseries resolver. when we have time, we can review the 0 vs null logic at the different timeseries sources and make sure logic looks sound.

matheus1lva:
Given that explanation and what we came to agree on, i've done the following changes:
1. PPS for timeseries returns null since it is plausible and fits much more into the context
2. TVL on the resolver is going to return null
3. TVL on yearn hooks i've added a check for the price to never be null (just in case).

I think we are good to go!

PR #229, opened 2025-04-03
feat: add forward APY calculation for vaults

matheus1lva:
Steps to test:

You can use `yarn ts-node packages/ingest/debug-apy.ts` or `bun packages/ingest/debug-apy.ts` to debug the implementation, by doing that you should see on the output table the `fapy` indicator.

Make sure to update some variables as well on your `.env`:

```
CRV_GAUGE_REGISTRY_URL=https://api.curve.fi/api/getAllGauges
CRV_POOLS_URL=https://api.curve.fi/api/getPools/all
YDAEMON_API="https://ydaemon.yearn.fi"

```
PS: By running `debug-apy`, it wont save on database because the output loaders happens on another queue.


.vscode/settings.json
murderteeth, line 1: whats this for?

config/crons.yaml
murderteeth, line 1: instead of implementing this as a new cron, lets do this like other timeseries, as a new timeseries hook. for example, https://github.com/yearn/kong/blob/main/packages/ingest/abis/yearn/2/vault/timeseries/apy/hook.ts

ie ingest/abis/yearn/2/vault/timeseries/fapy/hook.ts and move all the logic there

murderteeth (changes requested):

.vscode/settings.json
matheus1lva: ah i guess i've changed something on my editor and it added the config over here 🤡

config/crons.yaml
matheus1lva: oK!

packages/ingest/debug-fapy.ts
```diff
@@ -0,0 +1,54 @@
+// Load environment variables first
+import path from 'path'
```
matheus1lva: this can be used to test the fapy!

matheus1lva:
Since the beginning of this pr, curve has suffered an exploit on its dns, so `curve.fi` is poisoned and the api isnt working.

Update your `./.env` with:
```
CRV_GAUGE_REGISTRY_URL=https://api.curve.finance/api/getAllGauges
CRV_POOLS_URL=https://api.curve.finance/api/getPools/all
YDAEMON_API="https://ydaemon.yearn.fi"

```

murderteeth (changes requested):
- since it's been a while please rebase and lets test again
- move ingest/forward -> ingest/abis/yearn/lib/fapy
- why does the hook return the strategy and snapshot? we should only need to return whatever ydaemon is returning + any extras if you think they're interesting
- the test script takes a while to run for me, that are the bottlenecks
- port test script to a test in the test suite so we can run it all together

would also like to have a script that validates against ydaemon api, 
  - goal is to make sure we have approximately the same results as ydaemon for every vault
  - lives here packages/scripts/src/issue-229
  - open to ideas, but think it works like this: start kong connected to fork of livedb, run indexer to index fapy hook, in another terminal run validate script which fetches fapys from kong api and ydaemon api and checks how close they are to being the same

matheus1lva:
Changes incorporated:
1. refactor to make db calls more elegant and efficient
2. Handling of some numbers to be more in par with ydaemon
3. Added integration test which will query directly database if credentials provided.

packages/lib/types.ts
```diff
@@ -292,7 +292,7 @@ export const OutputSchema = z.object({
   value: z.any().transform(val => {
     const result = z.number().safeParse(val)
     if(result.success && isFinite(result.data)) return result.data
-    return undefined
+    return val
```
murderteeth: why return val instead of undefined here?

matheus1lva:
1. Unit tests added mocking soem calls
2. integration tests added

To call integration tests there's a npm script over to ingest that runs them!

matheus1lva:
I've reworked unit test for FAPY, now it has all data needed inserted into SQL that is needed!

Basides, i've decreased from 20s down to around 10s the startup of the test up to returning a value, by adding cache to the least updatable http call, besides leveraging `Promise.all`.

Did a few cleanups and apparently v2 calculation was not for fapy, so i took it out of the main flow for now!


I've also added to `ingest/package.json` a `validate:fapy` script, so we can run when needed!

Added fapy support on snapshot and on web2 graphql types!

Issue #232, opened 2025-04-04
Onboard Cove


murderteeth:
cove's vaults are built on yearn v3. so we can index under `ingest/abis/yearn/3`

cove has a registry here,
https://docs.cove.finance/security/contract-addresses

we can either index the cove registry or add their vaults to `config/manuals.yaml` for now.

when a new cove vault is detected lets add a `cove` flag to the vault's defaults. so a cove vault's default flags should look like

```

erc4626: true
v3: true
cove: true
yearn: undefined or false

```

Issue #225, opened 2025-03-25
Fill in missing APYs


murderteeth:
### what
We compute apy on external 4626 vaults here,
https://github.com/yearn/kong/blob/main/packages/ingest/abis/erc4626/timeseries/apy/hook.ts#L115

But our apy hook makes the assumption that a vault's decimal accuracy is the same as its asset. This is always true for yearn vaults, but not always true for external vaults. 

### so what
The result is that some of the index's external 4626 vaults incorrectly compute a 0% APY. For example, https://kalani.yearn.farm/erc4626/1/0xbEef047a543E45807105E51A8BBEFCc5950fcfBa

### now what
- Update kong's apy logic to not assume vault and asset decimals are the same
- Recompute historic APYs on effected external 4626 vaults

PR #233, opened 2025-04-04
Update manuals.yaml with Cove autocompounders

rossgalloway:
issue #232

### DIFF
--- a/config/manuals.yaml
+++ b/config/manuals.yaml
@@ -102,3 +102,181 @@ manuals:
       inceptBlock: 32945516,
       inceptTime: 1710521005
     }
+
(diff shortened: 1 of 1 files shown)


murderteeth (changes requested):
looking at these.. they are not v3-enough to be indexed as v3. missing pps for example causes lots of errors. so lets change `v3: true` to `v3: false` and remove `apiVersion: '3.0.0'` 

these will then get indexed as generic 4626 vaults. TVL looks ok, APYs not working because of https://github.com/yearn/kong/issues/225

image

rossgalloway:
stale

PR #253, opened 2025-08-08
Add webhooks for remote timeseries

murderteeth:
Webhooks are a way to outsource indexing logic to remote services.

includes --
- register of webhook subscriptions in config/subscriptions.yaml
- POSTs indexing requests to registered subscriptions every indexing cycle
- support for remote timeseries hooks 
- outputs are available from the gql timeseries query

also includes --
- a demonstration of how to integrate a webhook using nextjs at packages/web/app/api/webhook-healthcheck
- new testcases under config/testcases for dev testing


packages/web/app/api/webhook-healthcheck/route.ts
```diff
@@ -0,0 +1,76 @@
...
+  subscription: z.object({
+    id: z.string(),
+    url: z.string().url(),
+    abiPath: z.string(),
+    type: z.enum(['timeseries']),
+    label: z.string()
+  })
+})
+
+const OutputSchema = z.object({
```
matheus1lva, line 25: can we pull the main one over to `ingest`? So we have less duplication and the chances of this being stale!

matheus1lva (approved):

packages/web/app/api/webhook-healthcheck/route.ts
murderteeth: good catch! i did this on purpose actually. this way anyone can just copypaste from this route to start their own hook and only have a dependency on zod. the schemas could drift youre right, but these two in particular should be pretty stable

matheus1lva: Big brain moves

PR #262, opened 2025-10-13
Add katana yieldSplitter factory

murderteeth:


0xeye (commented):
Previous was mainnet but a test implementation, `0xfb277c7DfDa414aF824AF08c3596d6c28570347d` should be the replacement & not in an addition unless the previous is needed for a sync reason I'm unaware of

murderteeth:
> Previous was mainnet but a test implementation, `0xfb277c7DfDa414aF824AF08c3596d6c28570347d` should be the replacement & not in an addition unless the previous is needed for a sync reason I'm unaware of

main reason for keeping test imp is just provenance, not sync per se. ie, we have some objects tracked from the original factory in the database. keeping the the test factory in the config makes it clear how they got in the database. so instead of getting rid of it, i just added a comment "test implementation, obsolete"

image

0xeye (approved):
LGTM

PR #276, opened 2025-11-12
source metadata from ycms

matheus1lva:
source metadata from ycms using cdn style api

Usable for us to pull strategy name for some that currently do not have.

Ticket: BE-6


packages/lib/types.ts
```diff
@@ -230,29 +230,132 @@ export const DefaultRiskScore = RiskScoreSchema.parse({
 })
 
 export const TokenMetaSchema = z.object({
-  type: z.string(),
-  category: z.string(),
-  description: z.string(),
-  displayName: z.string(),
-  displaySymbol: z.string(),
-  icon: z.string()
+  // Legacy fields
```
murderteeth: where do these comments come from? (Legacy fields, Core fields, Display fields)

pls remove unless there's a good reason, otherwise will confuse

packages/lib/types.ts
```diff
@@ -230,29 +230,132 @@ export const DefaultRiskScore = RiskScoreSchema.parse({
...
 export const VaultMetaSchema = z.object({
-  displayName: z.string(),
-  displaySymbol: z.string(),
-  description: z.string(),
-  protocols: z.string().array().nullish()
+  chainId: z.number(),
+  address: AddressSchema,
+  name: z.string(),
+  registry: AddressSchema.optional(),
+  type: z.enum(['Yearn Vault', 'Experimental Yearn Vault', 'Automated Yearn Vault', 'Single Strategy', 'None']),
```
murderteeth: for all the enum fields, lets just use z.string in the kong\client schema. otherwise adding an enum value in cms requires updating kong too which seems unnecessary.

packages/web/app/api/gql/typeDefs/lib.ts
```diff
@@ -57,12 +57,20 @@ type RiskScoreDetails {
 }
 
 type TokenMeta {
+  # Legacy fields
```
murderteeth: same thing, why are these "Legacy" fields vs Display fields. unnecessary distinction, will just confuse agents

packages/web/app/api/gql/typeDefs/lib.ts
```diff
@@ -57,12 +57,20 @@ type RiskScoreDetails {
...
   type: String
-  category: String
-  description: String
+  icon: String
+  # Core fields from yCMS
+  chainId: Int
```
murderteeth: `chainId`, `address`, `name`, and `registry` are convenience fields used by cms. do not include these in the graphql typedef. this goes for all the typedefs, VaultMeta, TokenMeta, and StrategyMeta

murderteeth (changes requested):
i'd also like to start using a pr template in the kong repo. i PRd it against this branch to make sure you see it or have suggestions, https://github.com/yearn/kong/pull/277

otherwise looks good. i tested with `0x00C8a649C9837523ebb406Ceb17a6378Ab5C74cF`, USDC fluid lender

packages/lib/types.ts
matheus1lva: i had taken these out but i guess i've forgot here T.T

murderteeth (changes requested):
I see you fixed this for StrategyMetaSchema.protocol, good. There are still several enums left that should be converted to z.string(). Please check all the meta related schemas, VaultMetaSchema, StrategyMetaSchema, and TokenMetaSchema

for example
Image

packages/web/app/api/gql/typeDefs/vault.ts
```diff
@@ -1,11 +1,50 @@
...
+  isPoolTogether: Boolean
+  isCove: Boolean
+  isMorpho: Boolean
+  isKatana: Boolean
+  isPublicERC4626: Boolean
+}
+
 type VaultMeta {
+  name: String
+  registry: String
```
murderteeth: `VaultMeta.name` and `VaultMeta.registry` are convenience fields used internally by cms. remove from typedef

packages/web/app/api/gql/typeDefs/strategy.ts
```diff
@@ -2,6 +2,8 @@ import gql from 'graphql-tag'
 
 export default gql`
 type StrategyMeta {
+  name: String
```
murderteeth: `StrategyMeta.name` is a convenience field used internally by cms. remove from typedef

murderteeth (changes requested):

murderteeth (approved):

PR #278, opened 2025-11-15
fix: add vault address to strategy resolvers

matheus1lva:
> **Why this exists**

### Summary

During ydaemon integration of kong i've found out that during strategies query the vault field wasn't being populated correctly leading to unexpected behaviors.

image

### How to review

1. run graphql api against any populated db
2. run:
```
query Strategies($chainId: Int) {
  strategies(chainId: $chainId) {
    vault
    name
  }
}
```

you should see: 
image

### Test plan
- [ ] Manual: <steps, special cases, environment, etc>
- [ ] Automated (when applicable): <tests run / commands>

### Risk / impact
<Any risky areas (funds, auth, migrations, rate limits)?
Any feature flags, kill switches, or easy rollback steps?>

### DIFF
--- a/packages/web/app/api/gql/resolvers/strategies.ts
+++ b/packages/web/app/api/gql/resolvers/strategies.ts
@@ -11,18 +11,30 @@ const strategies = async (_: object, args: { chainId?: number, apiVersion?: stri
       thing.address,
       thing.defaults,
       snapshot.snapshot,
-      snapshot.hook
+      snapshot.hook,
+      (
+        SELECT vault_snapshot.address
+        FROM snapshot AS vault_snapshot
+        JOIN thing AS vault_thing
+          ON vault_thing.chain_id = vault_snapshot.chain_id
+          AND vault_thing.address = vault_snapshot.address
+        WHERE vault_thing.label = 'vault'
+          AND vault_thing.chain_id = thing.chain_id
+          AND vault_snapshot.snapshot->'get_default_queue' @> jsonb_build_array(thing.address)
+        LIMIT 1
+      ) AS vault
     FROM thing
     JOIN snapshot
       ON thing.chain_id = snapshot.chain_id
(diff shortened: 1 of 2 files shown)


packages/web/app/api/gql/resolvers/strategies.ts
```diff
@@ -11,18 +11,30 @@ const strategies = async (_: object, args: { chainId?: number, apiVersion?: stri
       thing.address,
       thing.defaults,
       snapshot.snapshot,
-      snapshot.hook
+      snapshot.hook,
+      (
```
murderteeth: inefficient sql, pls revert. instead the preferred method in kong is to ensure the vault field from the snapshot hook. this way the resolver queries dont have to change and maintain predictable performance. 

### root cause
v2 strategy contracts had a `vault` field. kong automatically picks up `vault` during v2 snapshots. v3 strategies do not have a vault field though, so `vault` doesn't exist on v3 snapshots. 

### solution
in the v3 strategy snapshot hook, return `vault: <address>` as one of the hook return type's fields. the field will automatically show up in the gql resolver when it spreads the hook, `...row.hook`, no changes needed.

murderteeth (changes requested):
Close without merge, expected behavior

### root cause
v2 strategy contracts have a `vault` field. kong automatically picks up `vault` during v2 snapshots. 

v3 strategies do not have a vault field though. this is because a v3 strategy can be a strategy in many different vaults. this is different than in v2. in v2 a strategy is only ever a strategy in a single vault.

This means `vault` doesn't exist on v3 snapshots which shows up as a null on the gql resolver. This is expected behavior.

### remarks
kong's gql api supports the **union** of v2 and v3 contract fields. it's common to encounter fields that only have a value for v2 objects or only for v3 objects. this is usually fine, but would be great to document better, field for field.

PR #281, opened 2025-11-25
add Edge-cached timeseries API

rossgalloway:
### Why this exists
Edge-cached timeseries API with scheduled refresh jobs to serve PPS/APY/TVL quickly from Redis, per plan.

### Summary

Built manifest-driven timeseries stack: labels, Postgres helpers, Keyv/Redis helper, 24h full refresh and 15m incremental scripts, edge route with component filtering and caching/CORS headers, plus scheduled GitHub Actions and env sample update. All code covered by new tests; build sanity check blocked only by sandbox FS permission for Next SWC cache.

### How to review

Look at packages/web/app/api/timeseries/* (manifest, db/redis helpers, refresh scripts, edge route + tests) first; workflows in .github/workflows/timeseries-*.yml, env addition in packages/web/.env.local.example. Ignore existing Node module-type warnings (harmless). Happy path: run Redis + Postgres env, run 24h then 15m scripts, curl edge route with/without components to see filtered arrays.

### Testing

- [x] Manual: Start Redis, set Postgres/Redis env, run bun packages/web/app/api/timeseries/refresh-24hr.ts then .../refresh-15min.ts, curl edge endpoints (default + explicit components) to confirm payload/headers.
- [x] Automated: bun --filter web test (all passing; Node module-type warning only). lint and build both passed.


rossgalloway:
I ran the verification and testing process locally and it seems snappy. Running Powerglove and having it read straight from localhost did not see a speedup, but that may be a problem with my setup.

packages/web/app/api/rest/timeseries/db.ts
```diff
@@ -0,0 +1,84 @@
...
+  chainId: number,
+  address: string,
+  label: string,
+): Promise<TimeseriesRow[]> {
+  const result = await db.query(
+    `
+    SELECT
+      chain_id AS "chainId",
+      address,
+      $3::text AS label,
```
matheus1lva:
```suggestion
      label,
```

packages/web/package.json
rossgalloway, line 17: Should I be using this keyv instead of the new keyv I added on line 26 below?

rossgalloway: Seems like no. I have kept it. I did revert the other changes made to the package.json though

murderteeth: - preferably we just use @keyv/redis
- can we keep those changes to package.json?  why were they reverted

murderteeth (changes requested):

packages/web/package.json
rossgalloway: claude said both keyv and keyv/reddis were needed.

The changes to the package.json that I reverted were made in this branch. I reverted them back to the default state in the main branch. I think it should be the same other than the keyv lib and the swc version.

murderteeth: > claude said both keyv and keyv/reddis were needed.

incorrect. you can remove the dep and apply this to packages/web/app/api/rest/timeseries/redis.ts

```

-import Keyv from 'keyv'
-import KeyvRedis from '@keyv/redis'
+import { createKeyv } from '@keyv/redis'

-export function createTimeseriesKeyv(store?: Keyv<any>['opts']['store']): Keyv {
-  if (store) {
-    return new Keyv({ store })
-  }
-
-  const redisUrl = process.env.GQL_CACHE_REDIS_URL
-  if (redisUrl) {
-    return new Keyv({ store: new KeyvRedis(redisUrl) })
-  }
-
-  return new Keyv()
+export function createTimeseriesKeyv() {
+  const redisUrl = process.env.GQL_CACHE_REDIS_URL || 'redis://localhost:6379'
+  return createKeyv(redisUrl)

```

> The changes to the package.json that I reverted were made in this branch. I reverted them back to the default state in the main branch. I think it should be the same other than the keyv lib and the swc version.

i made the latest package.json changes in this branch to clean up testing, https://github.com/yearn/kong/commit/698b74788a2d5b53e7e97bbc5e282e473b502d51. as i recall you were having a problem running tests under packages/web and changed the test runner to get it working. my changes cleaned that up. reverting back to main is fine, but i feel like we're going in a circle on this.

also there was a breaking change somewhere since last time i tested manually. `bun app/api/rest/timeseries/refresh-24hr.ts ` fails now:

```

Fetching vaults...
Found 1944 vaults (batch size: 10)
40 |     res = resolve
41 |     rej = reject
42 |   }).catch((err) => {
43 |     // replace the stack trace that leads to `TCP.onStreamRead` with one that leads back to the
44 |     // application that created the query
45 |     Error.captureStackTrace(err)
               ^
error: column "output.label" must appear in the GROUP BY clause or be used in an aggregate function
     length: 175,
   severity: "ERROR",
     detail: undefined,
       hint: undefined,
   position: "63",
 internalPosition: undefined,
 internalQuery: undefined,
      where: undefined,
     schema: undefined,
      table: undefined,
   dataType: undefined,
 constraint: undefined,
       file: "parse_agg.c",
    routine: "check_ungrouped_columns_walker",
       code: "42803"

```

murderteeth: ah it was be1a24f550317dd156862623761c345e4efb7c16. replacing the constant with the field in the select clause made the field required in the group by

rossgalloway: Ok, my bad on all of this. I have now:
- reverted the package.json to your updated version
- reverted `db.ts` to fix the problem with the 24 hr timeseries
- used the @keyv/redis import to get keyv working in redis.ts

Tested by running the different timeseries successfully, building and running the web package for kong, and verified the charts work correctly in powerglove.

Ready for re-review and hopefully merging

murderteeth (approved):
lgtm

PR #298, opened 2026-01-07
feat: add vault list edge api

matheus1lva:
### Summary

Add vault list edge api to be able to serve it quicker on the edge

### Test plan
- Terminal 1 - Start Redis:

```
docker run --rm -p 6379:6379 redis:latest
```

# Leave this running

- Terminal 2 - Configure and start web app:

```
# Configure packages/web/.env.local
# Use kong's webops read replica for postgres (not localhost)
cat > packages/web/.env.local <<EOF
POSTGRES_HOST=<webops-read-replica-host>
POSTGRES_DATABASE=<database-name>
POSTGRES_USER=<username>
POSTGRES_PASSWORD=<password>
EOF
```

- Start the web app
```
cd packages/web
bun dev
```

# Leave this running

- Terminal 3 - Run scripts and test:

```
cd packages/web

# Run refresh to populate cache
bun app/api/rest/lists/refresh.ts
```

# Test API endpoint
curl -v [http://localhost:3000/api/rest/lists/vaults/1](http://localhost:3000/api/rest/lists/vaults/1)

curl -v [http://localhost:3000/api/rest/lists/vaults](http://localhost:3000/api/rest/lists/vaults/1)

### Manual tested

image

image


murderteeth (changes requested):
i get a 500 when requesting `vaults/1`, but `vaults/` works fine

```

 ⨯ SyntaxError: "undefined" is not valid JSON
    at JSON.parse (<anonymous>)
    at GET (app/api/rest/lists/vaults/[chainId]/route.ts:43:39)
  41 |
  42 |   const wrapped = JSON.parse(cached as string)
> 43 |   const vaults: VaultListItem[] = JSON.parse(wrapped.value)
     |                                       ^
  44 |
  45 |   return NextResponse.json(vaults, {
  46 |     status: 200,
 GET /api/rest/lists/vaults/1 500 in 403ms

```

also same comment as on 296, make sure each step of test plan is in text in the pr so an agent can easily follow

matheus1lva:
Updated, i messed up some parsing. 

I also didnt notice from previous pr, that the links were not being rendered as links. I have an ext called Refined GitHub which renders for me, im gonna pay more attention :D

packages/web/app/api/rest/lists/vaults/route.ts
```diff
@@ -0,0 +1,59 @@
...
+  name: string
+}
+
+const corsHeaders = {
+  'access-control-allow-origin': '*',
+  'access-control-allow-methods': 'GET,OPTIONS',
+}
+
+export async function GET() {
+  const client = await createRedisClient()
```
murderteeth: This route uses raw Redis client with `client.keys()` while the per-chain route uses Keyv. This creates complexity where you need to manually handle Keyv's wrapper format (`JSON.parse(wrapped.value)`). Consider using Keyv consistently across both routes for simpler, more maintainable code.

packages/web/app/api/rest/lists/vaults/route.ts
```diff
@@ -0,0 +1,59 @@
+import { NextResponse } from 'next/server'
+import { createRedisClient } from '../redis'
+
+export const runtime = 'nodejs'
+
+type VaultListItem = {
```
murderteeth: This type is already defined in `db.ts` with Zod validation. Consider importing it from there instead of redefining it:
```typescript
import { VaultListItem } from '../db'
```
This ensures type consistency and avoids drift between definitions.

murderteeth (changes requested):
Thanks for adding the vault list API. A couple of suggestions to improve consistency and maintainability.

murderteeth (approved):
nm lgtm. lets merge and address those issues here https://github.com/yearn/kong/issues/300

packages/web/app/api/rest/lists/vaults/route.ts
matheus1lva: i mainly used this for regex grepping on keys from the same base prefix, its a bit difficulty to do the same with keyv, do you have any ideas?

matheus1lva: nvm, apparently can do with namespaces

PR #302, opened 2026-01-11
fix: force refresh-lists workflow scheduler re-registration

matheus1lva:
Add comment to trigger GitHub Actions to re-parse workflow and register cron schedule

ps: some people around internet says it works, ill give it a try


murderteeth (approved):
🤞

PR #318, opened 2026-01-28
Feat/extend vault murderteeth

murderteeth:
- instead of `netAPR` we should provide both `latestReportApr` and the full `performance` spread in strategy composition snapshots
-- this lets the client chose which performance measure is best
-- only v3 strategies will have `performance` because they are also vaults
- adds a `performance` spread to erc4626 snapshots for consistency

tested by targeting USDC-1 and USDC to USDS Depositor and indexing on a live db fork,

``` abis.local.yaml

  - abiPath: 'erc4626'
    sources: [
      { chainId: 1, address: '0x39c0aEc5738ED939876245224aFc7E09C8480a52', inceptBlock: 22440876 }
    ]

  - abiPath: 'yearn/3/vault'
    sources: [
      { chainId: 1, address: '0xBe53A109B494E5c9f97b9Cd39Fe969BE68BF6204', inceptBlock: 19419991 }
    ]

```

### DIFF
--- a/packages/ingest/abis/erc4626/snapshot/hook.ts
+++ b/packages/ingest/abis/erc4626/snapshot/hook.ts
@@ -25,6 +25,16 @@ export default async function process(chainId: number, address: `0x${string}`, d
     sparklines,
     tvl: sparklines.tvl[0],
     apy,
+    performance: {
+      estimated: undefined,
+      oracle: undefined,
+      historical: apy ? {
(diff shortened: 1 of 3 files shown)


matheus1lva:
Oh ok makes sense, given the strategy already reports performance. We provide the "raw" numbers, ie: performance, and one that represents the last reported by the stragey/vault.

Issue #320, opened 2026-01-29
Add per-strategy estimated APR to v2 strategy snapshots and vault composition


murderteeth:
## Description

Enhance the Kong indexer to consume per-strategy estimated APR data from the v2-estimated-apr-hook webhook. When the webhook outputs APR values for individual strategies (using strategy addresses), the v2 strategy snapshot hook should query and store this data in a `performance.estimated` object. The vault snapshot composition should then surface this strategy-level performance data.

## Context

**Current implementation:** The v2-estimated-apr-hook currently computes APR for each strategy individually but only outputs vault-level aggregated values (debt-ratio weighted). The Kong indexer only fetches estimated APR on the vault snapshot hook using the vault address, and v2 strategy snapshots have no `performance` object.

**Proposed/New system:** After yearn/v2-estimated-apr-hook#8 is implemented, the webhook will emit per-strategy outputs using strategy addresses with "raw" APR values (not debt-ratio weighted). The Kong indexer should:
1. V2 strategy snapshot hook queries `output` table by strategy address for estimated APR
2. V2 strategy snapshot hook returns a `performance` object: `{ estimated, oracle: undefined, historical: undefined }`
3. V2 vault composition fetches `hook->'performance'` from strategy snapshots and surfaces it

**Relevant references:**
- `packages/ingest/abis/yearn/2/strategy/snapshot/hook.ts` - V2 strategy snapshot hook (needs `performance` object)
- `packages/ingest/abis/yearn/2/vault/snapshot/hook.ts:55-110` - `getLatestEstimatedApr()` function to reuse/adapt
- `packages/ingest/abis/yearn/2/vault/snapshot/hook.ts:316-333` - `fetchStrategySnapshots()` needs to include `performance`
- `packages/ingest/abis/yearn/2/vault/snapshot/hook.ts:14-31` - `CompositionSchema` needs `performance` field
- `packages/ingest/abis/yearn/3/vault/snapshot/hook.ts:18-48` - V3 `CompositionSchema` with `performance` field (reference)
- `packages/lib/types.ts:443-460` - `EstimatedAprSchema` definition
- Commit 04bf5269c9411bbbc6ece26cde18506ab575ca82 - Example of adding estimated APR to v2 vault snapshots
- Commit 577bd20 - V3 vault composition with strategy performance

## Implementation Order

**This issue should be implemented FIRST**, before yearn/v2-estimated-apr-hook#8. This is safe because:
[...]

PR #321, opened 2026-01-29
Increase webhook max outputs from 20 to 100

matheus1lva:
### Summary
Increases the maximum allowed outputs for webhook subscriptions from 20 to 100. This change allows webhooks to handle larger result sets without hitting the artificial limit.

### Test plan
Trigger a webhook subscription that generates between 20-100 outputs and verify it processes successfully


murderteeth (approved):
this is good. we should be able to do everything in https://github.com/yearn/kong/issues/320 in one pr if you prefer. then move on to https://github.com/yearn/v2-estimated-apr-hook/issues/8

PR #322, opened 2026-01-29
feat: add isStakingEnabled support for vault metadata

matheus1lva:
### Summary
UIs need a metadata flag which through Cms we can serve the staking status, ie: if that's enabled or not seemlesly removing complexity from frontend.

### How to review
- Update cms adding flag for isStakingEnabled = false
- run indexer for that vault
- check snapshot.hook.meta.isStakingEnabled

### DIFF
--- a/packages/lib/types.ts
+++ b/packages/lib/types.ts
@@ -254,7 +254,7 @@ export const VaultMetaSchema = z.object({
   kind: z.string(),
   isRetired: z.boolean(),
   isHidden: z.boolean(),
-  isStakingEnabled: z.boolean(),
+  shouldDisableStaking: z.boolean(),
   isAggregator: z.boolean(),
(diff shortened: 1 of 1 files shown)


murderteeth (approved):
for new metadata to start showing up, we also have to update the metadata schema according to https://github.com/yearn/cms/blob/main/docs/how-to-manage-cms-metadata-schemas.md (it's simple)

matheus1lva:
> for new metadata to start showing up, we also have to update the metadata schema according to [yearn/cms@`main`/docs/how-to-manage-cms-metadata-schemas.md](https://github.com/yearn/cms/blob/main/docs/how-to-manage-cms-metadata-schemas.md) (it's simple)

https://github.com/yearn/cms/pull/101

Updated schema + added to a list of that should be disabled.

PR #328, opened 2026-01-30
fix: dont override meta on invalidation

matheus1lva:
### Summary
This PR improves the robustness of metadata fetching and ingestion. It modifies metadata fetching functions in `meta.ts` to return `undefined` instead of blank objects on failure, avoiding valid data override with empty strings. It also updates the snapshot ingestion loader to filter out `undefined` meta values, preserving existing database records when transient errors occur.

### How to review
Start with `packages/ingest/abis/yearn/lib/meta.ts` to see the logic change (returning `undefined` on catch).
Then check `packages/ingest/load/index.ts` to see how `upsertSnapshot` filters `undefined` values from `snapshot.hook` before merging.
Review `packages/ingest/abis/yearn/lib/meta.spec.ts` and `packages/ingest/abis/yearn/2/strategy/snapshot/hook.spec.ts` for test coverage.
Ignore local config changes if present.

### Test plan
- [x] Automated: `npx mocha abis/yearn/2/strategy/snapshot/hook.spec.ts` verify hook process handles validation failure gracefully.
- [x] Manual: Ingestion of valid/invalid vaults to ensure no data loss (verified via unit/integration tests).

### DIFF
--- a/packages/ingest/load/index.ts
+++ b/packages/ingest/load/index.ts
@@ -108,7 +108,11 @@ export async function upsertSnapshot(data: object) {
 
     const filteredHook = filterNullMeta(snapshot.hook)
 
-    snapshot.hook = { ...currentHook, ...filteredHook }
+    snapshot.hook = {
+      ...currentHook,
+      ...filteredHook,
+      meta: { ...currentHook.meta, ...filteredHook.meta }
+    }
     await upsert(snapshot, 'snapshot', 'chain_id, address', undefined, client)
 
     await client.query('COMMIT')
--- a/packages/ingest/load/index.ts
+++ b/packages/ingest/load/index.ts
@@ -87,7 +87,9 @@ export async function upsertEvmLog(data: object) {
 }
 
 function filterNullMeta(hook: Record<string, any>) {
-  return Object.fromEntries(Object.entries(hook).filter(([, v]) => v != null))
+  const meta = hook.meta
+  hook.meta = Object.fromEntries(Object.entries(meta).filter(([, v]) => v != null))
+  return hook
 }
 
 export async function upsertSnapshot(data: object) {
--- a/packages/ingest/load/index.ts
+++ b/packages/ingest/load/index.ts
@@ -104,7 +104,7 @@ export async function upsertSnapshot(data: object) {
     snapshot.hook = {
       ...currentHook,
       ...snapshot.hook,
-      meta: snapshot.hook.meta ? { ...currentHook.meta, ...snapshot.hook.meta } : currentHook.meta
+      meta: snapshot.hook.meta ? { ...currentHook.meta, ...JSON.parse(JSON.stringify(snapshot.hook.meta)) } : currentHook.meta
     }
     await upsert(snapshot, 'snapshot', 'chain_id, address', undefined, client)
 
--- a/packages/ingest/load/index.ts
+++ b/packages/ingest/load/index.ts
@@ -86,11 +86,6 @@ export async function upsertEvmLog(data: object) {
   }
 }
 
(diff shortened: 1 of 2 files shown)


murderteeth (changes requested):
## Review Summary

This PR partially addresses the bug where metadata parsing failures cause valid data to be overwritten with empty values. The current implementation handles individual item validation failures, but **does not fully solve the original issue**.

### What Works ✅
- Individual items that fail Zod schema validation are now skipped instead of crashing the entire `extractMetas` function
- Failed items are logged for debugging

### Critical Issues 🚨

**1. Outer functions still return empty objects on failure**

The catch blocks in `getVaultMeta`, `getStrategyMeta`, and `getTokenMeta` (lines 8-46) still return empty string values when errors occur:

```typescript
catch(error) {
  return {
    displayName: '',
    displaySymbol: '',
    description: '',
    protocols: []
  }
}
```

Per the bug report, these should return `undefined` so the loader can preserve existing data.

**2. Loader merge logic is unchanged**

The PR description mentions filtering `undefined` values in `load/index.ts`, but no changes were made to the loader. The current merge:
```typescript
snapshot.hook = { ...currentHook, ...snapshot.hook }
```
will still overwrite existing `meta` with empty strings when failures occur.

**3. PR description doesn't match actual changes**

- Description says: "return `undefined` instead of blank objects on failure" - not implemented
- Description says: "updates the snapshot ingestion loader to filter out `undefined` meta values" - not implemented  
- Description references tests - but only 1 file changed

matheus1lva:
wut, i forgot to push the load update T.T

packages/ingest/load/index.ts
```diff
@@ -99,7 +105,10 @@ export async function upsertSnapshot(data: object) {
     )) ?? { snapshot: {}, hook: {} }
 
     snapshot.snapshot = { ...currentSnapshot, ...snapshot.snapshot }
-    snapshot.hook = { ...currentHook, ...snapshot.hook }
+
+    const filteredHook = filterNullMeta(snapshot.hook)
+
+    snapshot.hook = { ...currentHook, ...filteredHook }
```
murderteeth: the result will still be null\undefined meta if getVaultMeta fails to pull fresh meta. all this change does is remove null properties from .meta, then save meta to the database without those fields.

murderteeth (changes requested):

packages/ingest/load/index.ts
matheus1lva: I was writting some other stuff, but then it clicked. I updated it!

murderteeth (changes requested):
The previous review concerns are addressed — meta functions now return `undefined` on failure, `extractMetas` handles individual parse errors gracefully, and the loader preserves existing meta when new meta is falsy.

One remaining edge case in the merge logic at `packages/ingest/load/index.ts:107`:

```typescript
meta: snapshot.hook.meta ? { ...currentHook.meta, ...snapshot.hook.meta } : currentHook.meta
```

When both `getVaultMeta` and `getTokenMeta` fail, vault hooks return `meta: { token: undefined }`. This object is truthy so the merge branch runs, but spreading `{ token: undefined }` overwrites `currentHook.meta.token`. When pg serializes to jsonb, `undefined` is dropped and the existing token data is lost.

Fix — strip `undefined` values before merging by round-tripping through JSON (no fidelity loss since this goes straight to jsonb):

```typescript
meta: snapshot.hook.meta
  ? { ...currentHook.meta, ...JSON.parse(JSON.stringify(snapshot.hook.meta)) }
  : currentHook.meta
```

murderteeth (approved):

PR #335, opened 2026-02-11
Switch webhooks to batch dispatch

matheus1lva:
### Summary
Replace per-vault webhook dispatch with a batch collector pattern. Instead of enqueuing one MQ job per vault per subscription, vaults are grouped by `subscriptionId:chainId` and dispatched as a single job per group. This reduces the number of HTTP calls to webhook subscribers and allows them to batch-optimize their processing (e.g. shared GQL queries, shared chain data fetching).

### How to review
1. Start with `packages/ingest/fanout/webhooks.ts` — the core change from `WebhooksFanout` worker to `WebhookCollector` class with `collect()` / `flush()`
2. Then `packages/ingest/fanout/abis.ts` — how the collector is used inline during ABI fanout
3. Then `packages/ingest/extract/webhook.ts` — DataSchema change to `vaults[]` and per-vault MAX_OUTPUTS
4. `packages/ingest/fanout/index.ts` and `packages/lib/mq.ts` are cleanup (removed webhook fanout worker and MQ job)

### Test plan
- [ ] Manual: Run `fanout abis` and verify webhook jobs are enqueued with batched `vaults[]` payloads
- [ ] Manual: Verify webhook subscribers receive the new batch format and respond correctly
- [ ] Automated: `bun --filter lib test` passes

### Risk / impact
- **Breaking change for webhook subscribers**: Subscribers must accept the new `vaults[]` array format instead of single `chainId`/`address`. The fapy-hook has already been updated in a companion PR.
- **MAX_OUTPUTS**: Now logs errors instead of throwing, so a misbehaving subscriber won't block ingestion.
- **Rollback**: Revert this commit and the subscriber changes.

Closes https://linear.app/yearn-webops/issue/BE-52/switch-webhooks-to-batch

### DIFF
--- a/packages/ingest/extract/webhook.ts
+++ b/packages/ingest/extract/webhook.ts
@@ -6,13 +6,11 @@ import { WebhookSubscription, WebhookSubscriptionSchema } from 'lib/subscription
 
 export const DataSchema = z.object({
   abiPath: z.string(),
+  chainId: z.number(),
   blockNumber: z.bigint({ coerce: true }),
   blockTime: z.bigint({ coerce: true }),
   subscription: WebhookSubscriptionSchema,
-  vaults: z.array(z.object({
-    chainId: z.number(),
-    address: zhexstring
-  }))
+  vaults: z.array(zhexstring)
 })
 
 export type Data = z.infer<typeof DataSchema>
--- a/packages/ingest/extract/webhook.ts
+++ b/packages/ingest/extract/webhook.ts
@@ -6,11 +6,13 @@ import { WebhookSubscription, WebhookSubscriptionSchema } from 'lib/subscription
 
 export const DataSchema = z.object({
   abiPath: z.string(),
-  chainId: z.number(),
-  address: zhexstring,
   blockNumber: z.bigint({ coerce: true }),
   blockTime: z.bigint({ coerce: true }),
-  subscription: WebhookSubscriptionSchema
+  subscription: WebhookSubscriptionSchema,
+  vaults: z.array(z.object({
+    chainId: z.number(),
+    address: zhexstring
+  }))
 })
 
 export type Data = z.infer<typeof DataSchema>
@@ -61,7 +63,7 @@ export class WebhookExtractor {
     await semaphore.acquire()
 
     try {
-      const label = `🔌 ${mq.job.extract.webhook.name} ${subscription.id} ${subscription.url} ${subscription.labels.join(', ')}`
+      const label = `🔌 ${mq.job.extract.webhook.name} ${subscription.id} ${subscription.url} ${subscription.labels.join(', ')} (${data.vaults.length} vaults)`
       console.time(label)
       const response = await fetchResponse(subscription, data)
       console.timeEnd(label)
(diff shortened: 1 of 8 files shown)


murderteeth (commented):
Spec compliance looks good — all 4 task groups are implemented correctly. Clean diff, tight collector pattern. Two minor things worth noting:

**1. Error log missing subscription ID** (`extract/webhook.ts`)
When a vault exceeds `MAX_OUTPUTS_PER_VAULT`, the error log doesn't include the subscription ID. Would help with production debugging.

**2. Label validation throws globally before per-vault filtering**
The label check on the flat `outputs` array throws before `Map.groupBy` runs. So one vault returning a bad label kills the entire batch — even if other vaults in the batch are fine. If the intent is per-vault resilience (skip bad vaults, keep good ones), the label check should move into the per-vault loop. If global fail-fast is intentional, this is fine as-is.

matheus1lva:
> Spec compliance looks good — all 4 task groups are implemented correctly. Clean diff, tight collector pattern. Two minor things worth noting:
> 
> **1. Error log missing subscription ID** (`extract/webhook.ts`) When a vault exceeds `MAX_OUTPUTS_PER_VAULT`, the error log doesn't include the subscription ID. Would help with production debugging.
> 
> **2. Label validation throws globally before per-vault filtering** The label check on the flat `outputs` array throws before `Map.groupBy` runs. So one vault returning a bad label kills the entire batch — even if other vaults in the batch are fine. If the intent is per-vault resilience (skip bad vaults, keep good ones), the label check should move into the per-vault loop. If global fail-fast is intentional, this is fine as-is.



> Spec compliance looks good — all 4 task groups are implemented correctly. Clean diff, tight collector pattern. Two minor things worth noting:
> 
> **1. Error log missing subscription ID** (`extract/webhook.ts`) When a vault exceeds `MAX_OUTPUTS_PER_VAULT`, the error log doesn't include the subscription ID. Would help with production debugging.
> 
> **2. Label validation throws globally before per-vault filtering** The label check on the flat `outputs` array throws before `Map.groupBy` runs. So one vault returning a bad label kills the entire batch — even if other vaults in the batch are fine. If the intent is per-vault resilience (skip bad vaults, keep good ones), the label check should move into the per-vault loop. If global fail-fast is intentional, this is fine as-is.

apparent 2 i tried to get around but made the same mistake. It was not intended. Fixed.

murderteeth (changes requested):
### Compile error on `make dev`

Running `make dev` with all defaults fails with a TypeScript compile error:

```
fanout/abis.ts:37:41 - error TS2345: Argument of type '{ chainId: number; address: `0x${string}`; inceptBlock: any; inceptTime: any; }' is not assignable to parameter of type '{ chainId: number; address: `0x${string}`; inceptBlock: bigint; skip: boolean; only: boolean; startBlock?: bigint | undefined; endBlock?: bigint | undefined; }'.
  Type '{ chainId: number; address: `0x${string}`; inceptBlock: any; inceptTime: any; }' is missing the following properties from type '{ chainId: number; address: `0x${string}`; inceptBlock: bigint; skip: boolean; only: boolean; startBlock?: bigint | undefined; endBlock?: bigint | undefined; }': skip, only

37           webhookCollector.collect(abi, _data.source)
                                           ~~~~~~~~~~~~
```

The thing-derived source object doesn't satisfy `SourceConfig` because the old code passed it through `mq.add` (which accepts `any`), bypassing type checks. The direct `collect()` call now surfaces the mismatch.

Please test changes end-to-end with `make dev` before requesting review.

matheus1lva:
I was also respecting the tests T.T, i've added `pretest` npm scripts on each package so when test is run there's also typecheck performed to add one more layer of check.

murderteeth (approved):
lgtm. but do not merge until we coordinate

PR #338, opened 2026-02-13
fix: revert sanitization on hook meta

matheus1lva:
Quickly revert it just for the sake of sanity.

### DIFF
--- a/packages/ingest/load/index.ts
+++ b/packages/ingest/load/index.ts
@@ -1,11 +1,11 @@
-import { z } from 'zod'
-import { mq, strider, types } from 'lib'
-import db, { firstRow, getTravelledStrides, toUpsertSql } from '../db'
-import { Processor } from 'lib/processor'
-import { PoolClient } from 'pg'
-import { OutputSchema, SnapshotSchema, ThingSchema, zhexstring } from 'lib/types'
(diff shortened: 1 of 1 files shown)


murderteeth (commented):
hold off on this. the reason pooltogether vaults aren't getting metadata is the same reason they have no apy or tvl. the vault filters in abis.yaml doesn't pick them up because they have `origin=ydaemon` and `apiVersion=~3.0.2` . these are older, atypical vaults. eg none of these vault filters will match them:
https://github.com/yearn/kong/blob/main/config/abis.yaml#L114
https://github.com/yearn/kong/blob/main/config/abis.yaml#L204
https://github.com/yearn/kong/blob/main/config/abis.yaml#L9

Image

instead we may need to implement a new path for these depending on how tvl and apy are computed.

matheus1lva:
But wait, if they are present in the db, that doesn't mean they matched?

murderteeth:
> But wait, if they are present in the db, that doesn't mean they matched?

i added them with a script at some point when matching kong vaults w ydaemon vaults. they didn't come from a factory that kong indexes, no

matheus1lva:
Ah then it makes sense!

PR #348, opened 2026-02-24
feat: add Sentry metrics for extract job tracking

matheus1lva:
## Summary

- Adds `@sentry/node` to `packages/lib` and initializes Sentry in `mq.ts` when `MQ_INVENTORY=true` and `SENTRY_DSN` is set
- Emits a `mq.extract_job_added` counter metric on every job added to the `extract-1` queue, with attributes: `jobName`, `address`, `chainId`, `fromBlock`, `toBlock`, `abiPath`
- Flushes Sentry on `beforeExit` to ensure all metrics are delivered
- Duplicate/anomaly detection is handled entirely in Sentry via dashboards and alerts — group by `jobName + address + chainId + fromBlock + toBlock` and alert where count > 1

## Files changed

- `packages/lib/package.json` — added `@sentry/node`
- `packages/lib/mq.ts` — Sentry init + metrics counter in `add()` + flush on `beforeExit`

### Test plan

- [ ] Go to [Sentry](https://sentry.io), open the **Kong** project and find in one of the section, ie: logs, metrics, etc, a code snippet to setup it and you will find the setry dsn secret, or create your own project and use it.
- [ ] Set up env vars in `packages/ingest/.env`
```bash
# add these two lines
MQ_INVENTORY=true
SENTRY_DSN=<your-dsn-from-sentry>
```

- [ ] Add a test source to `config/abis.local.yaml` (or use the existing one)
```yaml
cron:
  name: AbiFanout
  queue: fanout
  job: abis
  schedule: '*/15 * * * *'
  start: false

abis:
  - abiPath: 'yearn/3/vault'
    sources: [
      { chainId: 1, address: '0xAaaFEa48472f77563961Cdb53291DEDfB46F9040', inceptBlock: 24329199 }
    ]
```

- [ ] Start Redis and Postgres
```bash
make dev
```

- [ ] Navigate to the ingest pane in tmux and run a fanout
```
ctrl+b then arrow keys to navigate to ingest pane
select: fanout abis
```

- [ ] Wait for the fanout to dispatch extract jobs (watch the ingest logs)

- [ ] Go to Sentry > **Metrics**, filter by the **Kong** project, search for `mq.extract_job_added`
```
Expected: counter metric visible, grouped by jobName/address/chainId/fromBlock/toBlock using SUM as aggregation
```

- [ ] Verify duplicate detection: in the Sentry Metrics explorer, group by `address + chainId + fromBlock + toBlock` and check for any combination where the count > 1 — these are duplicate dispatches

image
you should see a table like this

### DIFF
--- a/packages/lib/mq.ts
+++ b/packages/lib/mq.ts
@@ -1,6 +1,11 @@
 import { Queue, Worker } from 'bullmq'
-import { Job } from './types'
+import fs from 'fs'
+import path from 'path'
 import chains from './chains'
+import { Job } from './types'
+
+const MQ_INVENTORY = process.env.MQ_INVENTORY === 'true'
+const MQ_INVENTORY_PATH = process.env.MQ_INVENTORY_PATH || path.resolve(process.cwd(), 'output/extract-1-inventory.csv')
 
 export const q = {
   fanout: 'fanout',
@@ -69,6 +74,30 @@ export function connect(queueName: string) {
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 export async function add(job: Job, data: any, options?: any) {
   const queue = job.bychain ? `${job.queue}-${data.chainId}` : job.queue
+  if (MQ_INVENTORY && queue === 'extract-1') {
+    try {
+      fs.mkdirSync(path.dirname(MQ_INVENTORY_PATH), { recursive: true })
+      if (!fs.existsSync(MQ_INVENTORY_PATH)) {
+        fs.writeFileSync(MQ_INVENTORY_PATH, 'jobName,abiPath,address,chainId,fromBlock,toBlock,outputLabel\n')
+      }
+      const csvEscape = (v: unknown) => {
+        const s = String(v ?? '')
+        return s.includes(',') || s.includes('"') ? `"${s.replace(/"/g, '""')}"` : s
+      }
+      const row = [
+        job.name,
+        data.abiPath ?? data.abi?.abiPath ?? '',
+        data.address ?? data.source?.address ?? '',
+        data.chainId ?? data.source?.chainId ?? '',
+        data.from ?? data.fromBlock ?? '',
+        data.to ?? data.toBlock ?? '',
+        data.outputLabel ?? ''
+      ].map(csvEscape).join(',')
+      fs.appendFileSync(MQ_INVENTORY_PATH, row + '\n')
+    } catch(error) {
+      console.error('📋', 'inventory write failed', error)
+    }
+  }
   if (!queues[queue]) { queues[queue] = connect(queue) }
   return await queues[queue].add(job.name, data, { priority: DEFAULT_PRIORITY, attempts: 1, ...options })
 }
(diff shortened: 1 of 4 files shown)


murderteeth (changes requested):
## Review: Sentry metrics for extract job tracking

The pivot from CSV logging to Sentry metrics is clean — no leftover CSV code. PR description accurately matches the final implementation. `Sentry.metrics.count()` is confirmed supported in SDK v10.25.0+ (currently in open beta).

Three changes requested below.

packages/lib/mq.ts
```diff
@@ -69,6 +77,18 @@ export function connect(queueName: string) {
...
+      attributes: {
+        jobName: job.name,
+        address: String(data.address ?? data.source?.address ?? ''),
+        chainId: String(data.chainId ?? data.source?.chainId ?? ''),
+        fromBlock: String(data.from ?? data.fromBlock ?? ''),
+        toBlock: String(data.to ?? data.toBlock ?? ''),
+        abiPath: String(data.abiPath ?? data.abi?.abiPath ?? '')
+      }
+    })
+  }
```
murderteeth: Since we're using Sentry, track all queues instead of just `extract-1`. Remove the `queue === 'extract-1'` filter and add `queue` as an attribute — you can still filter to any specific queue in Sentry dashboards.

```suggestion
  if (MQ_INVENTORY && SENTRY_DSN) {
    Sentry.metrics.count('mq.job_added', 1, {
      attributes: {
        queue,
        jobName: job.name,
        address: String(data.address ?? data.source?.address ?? ''),
        chainId: String(data.chainId ?? data.source?.chainId ?? ''),
        fromBlock: String(data.from ?? data.fromBlock ?? ''),
        toBlock: String(data.to ?? data.toBlock ?? ''),
        abiPath: String(data.abiPath ?? data.abi?.abiPath ?? '')
      }
    })
  }
```

Duplicate detection still works — group by `queue + jobName + address + chainId + fromBlock + toBlock` in Sentry.

packages/lib/mq.ts
```diff
@@ -1,6 +1,14 @@
 import { Queue, Worker } from 'bullmq'
-import { Job } from './types'
+import * as Sentry from '@sentry/node'
```
murderteeth: `@sentry/node` loads the full Sentry + OpenTelemetry instrumentation stack on import. Since this only runs when `MQ_INVENTORY` and `SENTRY_DSN` are set, consider a dynamic import to avoid the overhead in all other processes that touch `mq.ts`.

Something like:
```ts
let Sentry: typeof import('@sentry/node') | null = null

if (MQ_INVENTORY && SENTRY_DSN) {
  import('@sentry/node').then(s => {
    Sentry = s
    Sentry.init({ dsn: SENTRY_DSN })
  })
}
```

Then guard usage with `if (Sentry) { ... }`.

packages/lib/mq.ts
```diff
@@ -155,3 +175,9 @@ export function computeConcurrency(jobs: number, options: ConcurrencyOptions) {
 export async function down() {
   return Promise.all(Object.values(queues).map(async queue => queue.close()))
 }
+
+if (MQ_INVENTORY && SENTRY_DSN) {
+  process.on('beforeExit', async () => {
+    await Sentry.flush(5000)
+  })
+}
```
murderteeth: Node.js `beforeExit` does not await promises from async handlers — `Sentry.flush(5000)` may not complete before the process exits. Also, `beforeExit` doesn't fire on `SIGTERM`/`SIGINT` (the common shutdown signals).

Consider:
```ts
for (const signal of ['SIGTERM', 'SIGINT'] as const) {
  process.on(signal, async () => {
    await Sentry.flush(5000)
    process.exit(0)
  })
}
```

Or call `Sentry.flush()` explicitly in the existing `down()` function since that's already the shutdown path.

murderteeth (commented):
Line comments with suggested changes attached.

murderteeth (approved):
lgtm. i should have been more clear on when to flush sentry so i went ahead and added a commit for it

matheus1lva:
> lgtm. i should have been more clear on when to flush sentry so i went ahead and added a commit for it

Fair!

PR #359, opened 2026-03-09
fix: use Redis MSET instead of Keyv setMany for cache writes

matheus1lva:
## Summary
- keyv had outdated info when #343  was implemented, implying `setMany` was using mset or any more efficient command under the hood, which after another lookup again today, it's not there anymore, and through source code check, it does multiple sets. Which goes against what we want.
- Replace Keyv's `setMany()` (MULTI + N SETs + EXEC = N+2 commands) with Redis `MSET` (1 command) for all REST cache refresh scripts
- Access the underlying `@redis/client` from Keyv's store to call `mSet()` directly
- Values are serialized as `JSON.stringify({ value })` to stay compatible with `keyv.get()` reads in route handlers (which extract `.value` from deserialized data)
- No changes to route handlers — reads still go through `keyv.get()`

### Why

On Upstash pay-as-you-go, each command in a MULTI/EXEC transaction is billed individually. Keyv's `setMany(N)` produces N+2 billed commands (MULTI + N SETs + EXEC). `MSET` is a single Redis command regardless of how many keys it sets — billed as **1 command** on Upstash.

### Risks

- None — key count didn't change
- Value format didn't change (`JSON.stringify({ value })` matches Keyv's internal serialization)

image

## Test plan

- [ ] Start Redis
```
docker compose up -d redis
```

- [ ] Set up `packages/web/.env` with local Redis + read-only Postgres
```env
POSTGRES_HOST=ep-raspy-art-a5g9bur0-pooler.us-east-2.aws.neon.tech
POSTGRES_DATABASE=neondb
POSTGRES_USER=dev
POSTGRES_PASSWORD=<password>
POSTGRES_SSL=true
POSTGRES_PORT=5432
```

- [ ] Run all refresh scripts to populate the cache
```
cd packages/web && source .env && bun run app/api/rest/list/refresh.ts
```
Expected: `✓ Completed: 2049 vaults cached across 11 chains`

```
cd packages/web && source .env && bun run app/api/rest/snapshot/refresh-snapshot.ts
```
Expected: `✓ Completed: <N> vaults processed`

```
cd packages/web && source .env && bun run app/api/rest/reports/refresh.ts
```
Expected: `✓ Completed: <N> vaults processed`

```
cd packages/web && source .env && bun run app/api/rest/timeseries/refresh-timeseries.ts
```
Expected: `✓ Completed: <N> vaults processed`

- [ ] Start the web server
```
cd packages/web && source .env && bun run dev
```

- [ ] Verify list endpoints
```
curl -s http://localhost:3000/api/rest/list/vaults | jq 'length'
```
Expected: `2049`

```
curl -s http://localhost:3000/api/rest/list/vaults/1 | jq 'length'
```
Expected: `1210`

- [ ] Verify snapshot endpoint
```
curl -s http://localhost:3000/api/rest/snapshot/1/0xa3931d71877C0E7a3148CB7Eb4463524FEc27fbD | jq 'keys'
```
Expected:
```json
["address","apy","asset","blockNumber","blockTime","chainId","decimals","erc4626","inceptBlock","inceptTime","name","performance","pricePerShare","sparklines","symbol","totalAssets","totalSupply","tvl"]
```

- [ ] Verify timeseries endpoint
```
curl -s http://localhost:3000/api/rest/timeseries/tvl/1/0xa3931d71877C0E7a3148CB7Eb4463524FEc27fbD | jq '.[0:2]'
```
Expected:
```json
[
  {"time": 1725408000, "component": "tvl", "value": "0"},
  {"time": 1725494400, "component": "tvl", "value": "0"}
]
```

- [ ] Cleanup
```
docker compose down redis
```

### DIFF
--- a/packages/web/app/api/rest/cache.ts
+++ b/packages/web/app/api/rest/cache.ts
@@ -1,8 +1,11 @@
 import { createKeyv } from '@keyv/redis'
+import { createClient } from '@redis/client'
+
+const writeClient = createClient({ url: process.env.REST_CACHE_REDIS_URL || 'redis://localhost:6379' })
 
 const keyv = createKeyv(process.env.REST_CACHE_REDIS_URL || 'redis://localhost:6379')
 
-export function createKeyvClient() {
+export function getKeyvClient() {
   return keyv
 }
 
@@ -10,14 +13,14 @@ export function createKeyvClient() {
  * Batch write using Redis MSET (1 command) instead of Keyv's
  * setMany which uses MULTI + N SETs + EXEC (N+2 commands).
  */
+
 export async function cacheMSet(pairs: Array<[string, string]>): Promise<void> {
   if (pairs.length === 0) return
-  const store = keyv.store as { client: { mSet: (pairs: Array<[string, string]>, options?: unknown) => Promise<unknown>; isOpen: boolean; connect: () => Promise<unknown> } }
-  const client = store.client
-  if (!client.isOpen) await client.connect()
-  await client.mSet(pairs)
+  if (!writeClient.isOpen) await writeClient.connect()
+  await writeClient.mSet(pairs)
 }
 
 export async function disconnect(): Promise<void> {
+  await writeClient.disconnect()
   await keyv.disconnect()
 }
--- a/packages/web/app/api/rest/cache.ts
+++ b/packages/web/app/api/rest/cache.ts
@@ -1,7 +1,23 @@
 import { createKeyv } from '@keyv/redis'
 
+const keyv = createKeyv(process.env.REST_CACHE_REDIS_URL || 'redis://localhost:6379')
+
 export function createKeyvClient() {
-  const redisUrl = process.env.REST_CACHE_REDIS_URL || 'redis://localhost:6379'
-  return createKeyv(redisUrl)
+  return keyv
 }
 
+/**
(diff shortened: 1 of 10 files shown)


murderteeth (changes requested):
### Requested changes

1. **Use an isolated `@redis/client` instance for writes instead of reaching through Keyv internals.** The current `cacheMSet` drills into `keyv.store.client` via type assertion to access the raw Redis client's `mSet`. This is fragile — a `@keyv/redis` version bump could break it silently at runtime. Since refresh scripts only write (never read), create a dedicated write client:

```ts
import { createClient } from '@redis/client'

const writeClient = createClient({ url: process.env.REST_CACHE_REDIS_URL || 'redis://localhost:6379' })

export async function cacheMSet(pairs: Array<[string, string]>): Promise<void> {
  if (pairs.length === 0) return
  if (!writeClient.isOpen) await writeClient.connect()
  await writeClient.mSet(pairs)
}
```

`@redis/client` is already a transitive dependency via `@keyv/redis`, so no new dep needed. Route handlers continue reading through Keyv as-is.

2. **Rename `createKeyvClient` to reflect it's now a singleton accessor** (e.g. `getKeyvClient` or export `keyv` directly). Not blocking.

murderteeth (approved):
LGTM. Thank you for the detailed manual test plan.

PR #364, opened 2026-03-17
Fix database query performance and connection timeouts

murderteeth:
### Summary
Fixes database connection timeouts and high CPU usage on Neon caused by `be152e7` introducing `LOWER()` calls in address queries (killing index usage on 22M row output table) and N*3 per-strategy DB queries in the snapshot hook. Also re-applies the deterministic jobId dedup from #358, optimizes the system probe, and adds address checksumming to `OutputSchema`.

### How to review
1. **`apy-apr.ts`** — all `LOWER(address) = LOWER($2)` replaced with `address = $2`. These were added in `be152e7` and prevented use of `idx_output_chain_id_address`.
2. **`snapshot/hook.ts`** — `fetchStrategyPerformance` rewritten from N*3 individual queries to a single batch CTE query. `fetchStrategySnapshots` LOWER() also removed.
3. **`lib/types.ts`** — `OutputSchema.address` changed from `zhexstring` to `EvmAddressSchema` so webhook-ingested addresses get checksummed via `getAddress()`.
4. **`probe/index.ts`** — `count(*)` replaced with TimescaleDB `approximate_row_count()`, event counts cached for 5 min, schedule changed from 2s to 10s.
5. **`checksum-output-addresses.ts`** — one-time script to fix 74 existing lowercase rows in the output table.

### Test plan
- [x] Manual: ran full indexer against forked prod DB, confirmed snapshot times dropped from 2-4 min to 11-30s
- [x] Manual: ran checksum script against forked prod DB, handles duplicate key conflicts
- [x] Manual: verify dashboard probe data (output/evmlog counts, event breakdown) displays correctly after deploy

### Release procedure
1. Ensure ABI fanout schedule is disabled on prod
2. Merge this PR
3. Run `checksum-output-addresses.ts` against prod DB
4. Manually run ABI fanout on prod, wait for finish
5. Enable scheduled ABI fanout on prod

### Risk / impact
- Removing `LOWER()` means queries rely on addresses being consistently checksummed in the DB. We verified only 74 out of 22M output rows were lowercase (all from the yvusd webhook). The `OutputSchema` fix prevents future occurrences.
- The probe `approximate_row_count()` is an estimate, not exact. Dashboard counts may differ slightly from true values.
- Event counts are cached for 5 minutes — dashboard will show stale counts briefly after new event types appear.

### DIFF
--- a/packages/scripts/src/checksum-output-addresses.ts
+++ b/packages/scripts/src/checksum-output-addresses.ts
@@ -1,0 +1,69 @@
+/**
+ * Checksum lowercase addresses in the output table.
+ *
+ * Finds output rows where the address is lowercase instead of EIP-55 checksummed,
+ * and updates them to use the correct checksummed format.
+ */
+
+import 'dotenv/config'
+import { Pool } from 'pg'
+import { getAddress } from 'viem'
+
+async function main() {
+  const db = new Pool({
+    host: process.env.POSTGRES_HOST ?? 'localhost',
+    port: (process.env.POSTGRES_PORT ?? 5432) as number,
+    ssl: (process.env.POSTGRES_SSL ?? false)
+      ? (process.env.POSTGRES_SSL_REJECT_UNAUTHORIZED ?? true)
+        ? true
+        : { rejectUnauthorized: false }
+      : false,
+    database: process.env.POSTGRES_DATABASE ?? 'user',
+    user: process.env.POSTGRES_USER ?? 'user',
+    password: process.env.POSTGRES_PASSWORD ?? 'password',
+  })
+
+  const { rows: lowercased } = await db.query(`
+    SELECT DISTINCT address FROM output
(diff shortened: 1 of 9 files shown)


packages/scripts/src/checksum-output-addresses.ts
```diff
@@ -0,0 +1,69 @@
...
+
+  const { rows: lowercased } = await db.query(`
+    SELECT DISTINCT address FROM output
+    WHERE address ~ '[a-f]' AND address !~ '[A-F]'
+  `)
+
+  console.log(`Found ${lowercased.length} lowercase addresses`)
+
+  for (const { address } of lowercased) {
+    const checksummed = getAddress(address)
```
matheus1lva, line 34: Oh I didn't know getAddress from viem was actually making it eip compliant. That makes much easier rn! Thanks for bringuing this!

matheus1lva (approved):

PR #366, opened 2026-03-17
feat: add netApr and netApy components to apr-oracle timeseries

matheus1lva:
## Summary

- Add `netApr` and `netApy` components to the apr-oracle timeseries, computed by applying vault fee deductions to the gross oracle APR
- Extract `computeNetApr` and `computeApy` into shared utilities in `lib/apy.ts` so all consumers (timeseries hook, snapshot hook, backfill script) use the same formulas
- Add backfill script (`backfill-apr-oracle-netapr.ts`) to retroactively populate `netApr`/`netApy` for existing timeseries rows
- Add copy script (`copy-output-fork-to-main.ts`) to transfer backfilled data from the fork DB to the main DB

### Net APR formula
```
netApr = (grossApr - managementFee) * (1 - performanceFee)
```

### Net APY formula (weekly compounding)
```
netApy = (1 + netApr / 52)^52 - 1
```

## Changes
- `packages/ingest/abis/yearn/lib/apy.ts` — new shared `computeNetApr` and `computeApy` exports
- `packages/ingest/abis/yearn/3/vault/timeseries/apr-oracle/hook.ts` — emits `netApr` and `netApy` outputs using shared functions
- `packages/ingest/abis/yearn/3/vault/snapshot/hook.ts` — uses shared functions instead of inline formulas
- `packages/scripts/src/backfill-apr-oracle-netapr.ts` — backfill script for historical data (already ran against fork DB)
- `packages/scripts/src/copy-output-fork-to-main.ts` — copies backfilled `netApr`/`netApy` rows from fork DB to main DB
- Tests added/moved to `lib/apy.spec.ts`

### Unit tests

- [ ] Run ingest tests to verify `computeNetApr` and `computeApy`
```
bun --filter ingest test
```
Expected: all tests pass, including `computeNetApr` and `computeApy` specs in `lib/apy.spec.ts`

### Integration test (timeseries hook + snapshot hook)

```
make dev
```

- [ ] Create `config/abis.local.yaml` with a test vault
```yaml
cron:
  name: AbiFanout
  queue: fanout
  job: abis
  schedule: '*/15 * * * *'
  start: false

abis:
  - abiPath: 'yearn/3/vault'
    sources: [
      { chainId: 1, address: '0xBe53A109B494E5c9f97b9Cd39Fe969BE68BF6204', inceptBlock: 19419991 }
    ]

  - abiPath: 'yearn/3/strategy'
    things: {
      label: 'strategy',
      filter: [{ field: 'apiVersion', op: '>=', value: '3.0.0' }]
    }
```

- [ ] Create `config/manuals.local.yaml` so the vault is registered as a thing
```yaml
manuals:
  - chainId: 1
    address: '0xBe53A109B494E5c9f97b9Cd39Fe969BE68BF6204'
    label: 'vault'
    defaults: {
      erc4626: true,
      v3: true,
      yearn: true,
      origin: 'yearn',
      apiVersion: '3.0.2',
      registry: '0xff31A1B020c868F6eA3f61Eb953344920EeCA3af',
      asset: '0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48',
      decimals: 6,
      inceptBlock: 19419991,
      inceptTime: 1710258455
    }
```

- [ ] From the terminal UI, run `fanout abis` (run 2-3 times for full initialization)

- [ ] Verify `netApr` and `netApy` timeseries outputs were created
```
psql $POSTGRES_DATABASE -c "
  SELECT component, value, block_number
  FROM output
  WHERE label = 'apr-oracle'
    AND component IN ('netApr', 'netApy')
    AND chain_id = 1
    AND address = '0xBe53A109B494E5c9f97b9Cd39Fe969BE68BF6204'
  ORDER BY block_number DESC
  LIMIT 10;
"
```
Expected: rows with `netApr` and `netApy` components, values between 0 and 1

- [ ] Verify snapshot includes `netAPR` and `netAPY` in the oracle performance block
```
psql $POSTGRES_DATABASE -c "
  SELECT hook->'performance'->'oracle' AS oracle
  FROM snapshot
  WHERE chain_id = 1
    AND address = '0xBe53A109B494E5c9f97b9Cd39Fe969BE68BF6204'
  LIMIT 1;
"
```
Expected: JSON with `apr`, `netAPR`, `apy`, `netAPY` fields, all numeric

## Deployment procedure

The backfill was already run against a Neon fork DB (~787k output rows for netApr + netApy). After merge:

### Step 1 — Deploy

Deploy the new code to production. The indexer will start emitting `netApr`/`netApy` for new timeseries data going forward.

### Step 2 — Copy historical data from fork to main

Go to Render shell and run:
```bash
bun run packages/scripts/src/copy-output-fork-to-main.ts
```
Requires `FORK_POSTGRES_*` env vars (`FORK_POSTGRES_HOST`, `FORK_POSTGRES_DATABASE`, `FORK_POSTGRES_USER`, `FORK_POSTGRES_PASSWORD`, `FORK_POSTGRES_SSL`) pointing to the Neon fork DB.

### Step 3 — Verify
```sql
SELECT component, COUNT(*), MIN(series_time), MAX(series_time)
FROM output
WHERE component IN ('netApr', 'netApy')
GROUP BY component;
```
Expected: rows for both `netApr` and `netApy` spanning the full historical range.

### FAQ

**Do we need to stop the indexer?**
No. The copy script and the indexer write to different time ranges — the script backfills historical data while the indexer writes new data going forward. Both use upserts (`ON CONFLICT ... DO UPDATE`), so even if there's overlap at the boundary there's no conflict or data corruption.

**How long does the copy take?**
The script copies ~787k rows in read batches of 10k and write batches of 1k (concurrency 10). It's a straightforward DB-to-DB copy (no RPC calls or computation), so it should complete in a few minutes depending on network latency between Render and Neon.

**What if the copy fails mid-way?**
Safe to re-run. All writes are upserts, so already-copied rows are just overwritten with the same values. The script will re-read from the beginning but the writes are idempotent.

**Can we run step 1 and step 2 in any order?**
Yes. Deploy first is preferred so the indexer starts covering new data immediately, but the copy script is independent — it reads from the fork DB and writes to main DB regardless of what the indexer is doing.

### DIFF
--- a/packages/ingest/abis/yearn/3/vault/snapshot/hook.ts
+++ b/packages/ingest/abis/yearn/3/vault/snapshot/hook.ts
@@ -24,6 +24,8 @@ export const CompositionSchema = z.object({
   performance: z.object({
     estimated: EstimatedAprSchema.nullish(),
     oracle: z.object({
+      netAPR: z.number().nullish(),
+      netAPY: z.number().nullish(),
       apr: z.number().nullish(),
       apy: z.number().nullish()
     }).nullish(),
--- a/packages/ingest/abis/yearn/3/vault/snapshot/hook.ts
+++ b/packages/ingest/abis/yearn/3/vault/snapshot/hook.ts
@@ -10,6 +10,7 @@ import { getLatestApy, getLatestEstimatedAprV3, getLatestOracleApr } from '../..
 import { fetchErc20PriceUsd } from '../../../../../prices'
 import { rpcs } from '../../../../../rpcs'
 import * as things from '../../../../../things'
+import { computeApy, computeNetApr } from '../../../lib/apy'
 import { fetchOrExtractErc20 } from '../../../lib'
 import { getStrategyMeta, getTokenMeta, getVaultMeta } from '../../../lib/meta'
 import { getRiskScore } from '../../../lib/risk'
@@ -139,6 +140,10 @@ export default async function process(chainId: number, address: `0x${string}`, d
     LIMIT 1
   `, [chainId, address])
 
+  const oracleNetApr = oracleApr != null
+    ? computeNetApr(oracleApr, { management: fees.managementFee / 10_000, performance: fees.performanceFee / 10_000 })
+    : undefined
+
   return {
     asset, strategies, allocator, roles, debts, composition, fees,
     risk, meta: { ...meta, token },
@@ -149,13 +154,9 @@ export default async function process(chainId: number, address: `0x${string}`, d
       estimated: estimatedApr ?? undefined,
       oracle: {
         apr: oracleApr,
-        netAPR: oracleApr != null
-          ? oracleApr * (1 - fees.performanceFee / 10_000) - fees.managementFee / 10_000
-          : undefined,
+        netAPR: oracleNetApr,
         apy: oracleApy,
-        netAPY: oracleApy != null
-          ? oracleApy * (1 - fees.performanceFee / 10_000) - fees.managementFee / 10_000
-          : undefined,
+        netAPY: oracleNetApr != null ? computeApy(oracleNetApr) : undefined,
       },
       historical: apy ? {
         net: apy.net,
--- a/packages/ingest/abis/yearn/3/vault/snapshot/hook.ts
+++ b/packages/ingest/abis/yearn/3/vault/snapshot/hook.ts
@@ -152,7 +152,10 @@ export default async function process(chainId: number, address: `0x${string}`, d
         netAPR: oracleApr != null
           ? oracleApr * (1 - fees.performanceFee / 10_000) - fees.managementFee / 10_000
           : undefined,
-        apy: oracleApy
+        apy: oracleApy,
+        netAPY: oracleApy != null
+          ? oracleApy * (1 - fees.performanceFee / 10_000) - fees.managementFee / 10_000
+          : undefined,
       },
       historical: apy ? {
         net: apy.net,
--- a/packages/scripts/src/backfill-apr-oracle-netapr.ts
+++ b/packages/scripts/src/backfill-apr-oracle-netapr.ts
@@ -26,6 +26,7 @@ type TimeseriesCandidate = {
 type FeeConfig = { management: number; performance: number }
 type FeeConfigBps = { management: number; performance: number }
 type FeeSegment = { blockNumber: bigint } & FeeConfigBps
+type AccountantSegment = { blockNumber: bigint; accountant: `0x${string}` }
 
 type StrategyChangeEvent = { blockNumber: bigint; strategy: `0x${string}`; type: 'add' | 'revoke' }
 type DebtSnapshot = { blockNumber: bigint; currentDebt: bigint }
@@ -41,6 +42,9 @@ const CUSTOM_FEE_SELECTOR = toEventSelector(
 )
 const REMOVED_CUSTOM_FEE_SELECTOR = toEventSelector(
   'event RemovedCustomFeeConfig(address indexed vault, address indexed strategy)'
+)
+const UPDATE_ACCOUNTANT_SELECTOR = toEventSelector(
+  'event UpdateAccountant(address indexed accountant)'
 )
 
 function parseArgs(argv: string[]): Args {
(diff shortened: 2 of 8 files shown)


rossgalloway:
Hey, checking in on this guy.

Also wondering if we need to do this for the raw performance.oracle values? I don't remember if we discussed whether we were removing the fees from those values as well. May be good to have net values for them too.

matheus1lva:
> Hey, checking in on this guy.
> 
> Also wondering if we need to do this for the raw performance.oracle values? I don't remember if we discussed whether we were removing the fees from those values as well. May be good to have net values for them too.

https://github.com/yearn/kong/blob/main/packages/ingest/abis/yearn/3/vault/snapshot/hook.ts#L150 we do it here

murderteeth (changes requested):
This pr will only add net components to apr oracle timeseries records going forward. What about previous timeseries records?

rossgalloway:
> This pr will only add net components to apr oracle timeseries records going forward. What about previous timeseries records?

yes, +1 for getting historical data backfilled to correctly account for fees

packages/scripts/src/backfill-apr-oracle-netapr.ts
murderteeth, line 1: Filename has a double `.ts.ts` extension — should be `backfill-apr-oracle-netapr.ts`.

packages/scripts/src/backfill-apr-oracle-netapr.ts
```diff
@@ -0,0 +1,268 @@
...
+  const next = apr * (1 - fees.performance) - fees.management
+  return Number.isFinite(next) ? next : 0
+}
+
+function computeApy(apr: number): number {
+  const next = (1 + apr / 52) ** 52 - 1
+  return Number.isFinite(next) ? next : 0
+}
+
+async function getTimeseriesCandidates(args: Args): Promise<TimeseriesCandidate[]> {
```
murderteeth, line 77: This formula (`apr * (1 - fees.performance) - fees.management`) doesn't match the timeseries hook's `computeNetApr`: `(grossApr - fees.management) * (1 - fees.performance)`. Consider importing and reusing `computeNetApr` directly from the hook to guarantee consistency.

packages/ingest/abis/yearn/3/vault/snapshot/hook.ts
```diff
@@ -150,7 +152,10 @@ export default async function process(chainId: number, address: `0x${string}`, d
         netAPR: oracleApr != null
           ? oracleApr * (1 - fees.performanceFee / 10_000) - fees.managementFee / 10_000
           : undefined,
-        apy: oracleApy
+        apy: oracleApy,
+        netAPY: oracleApy != null
+          ? oracleApy * (1 - fees.performanceFee / 10_000) - fees.managementFee / 10_000
+          : undefined,
```
murderteeth: Two issues here:
1. The net APR formula `oracleApr * (1 - perf/10_000) - mgmt/10_000` differs from the timeseries hook's correct formula `(oracleApr - mgmt) * (1 - perf)`. These should match.
2. `netAPY` applies the same linear fee deduction to an already-compounded APY value. Instead, compute netAPR first and then compound: `(1 + netApr/52)^52 - 1`.

murderteeth (changes requested):
## Review

### Critical: Fee formula mismatch across three locations

The timeseries hook, snapshot hook, and backfill script each use a different net APR formula:

| Location | Formula |
|---|---|
| Timeseries hook `computeNetApr` | `(grossApr - mgmt) * (1 - perf)` |
| Snapshot hook (lines 152-153, 156-157) | `grossApr * (1 - perf) - mgmt` |
| Backfill script `computeFeeAdjustedApr` | `apr * (1 - perf) - mgmt` |

The timeseries hook is correct — confirmed by the inverse formula in `packages/ingest/abis/yearn/lib/apy.ts:204-206`:
```ts
grossApr = netApr / (1 - fees.performance) + fees.management
```
Inverting: `netApr = (grossApr - mgmt) * (1 - perf)`.

The snapshot hook and backfill script apply the performance fee to the full gross APR (including the management fee portion), which overcharges the user. The snapshot hook's `netAPY` has an additional issue: it applies a linear fee deduction to an already-compounded value instead of computing netAPR first and then compounding via `(1 + netApr/52)^52 - 1`.

### High: Backfill script uses current fees for all historical rows

The backfill script loads one current snapshot per vault (`getV3VaultSnapshots` at line 132) and caches a single fee result per vault (`feesByVault` at line 177). This is not historical state — `snapshot` is keyed by `(chain_id, address)` and each update merges into that single row. The live hook correctly resolves fees at each row's `blockNumber` via `extractFees__v3(chainId, address, strategies, blockNumber)`. Any vault whose fee config changed over time will get incorrect historical `netApr`/`netApy` from the backfill.

### Notes on the PR body / test plan

- The test plan is missing a `manuals.local.yaml` for `0xBe53A109B494E5c9f97b9Cd39Fe969BE68BF6204` — the provided config only has `abis.local.yaml`.
- How was the backfill script tested?
- How long does the backfill take? Should we disable the indexer, run the script, then turn the indexer back on? Please provide a procedure.

matheus1lva:
Updated, im generating now metrics to also add to the pr!

packages/scripts/src/backfill-apr-oracle-netapr.ts
```diff
@@ -0,0 +1,203 @@
...
+  let inserted = 0
+
+  for (const [index, candidate] of candidates.entries()) {
+    const cacheKey = `${candidate.chainId}:${candidate.address}:${candidate.blockNumber}`
+    let fees = feeCache.get(cacheKey)
+
+    if (!fees) {
+      try {
+        const strategies = await projectStrategies(candidate.chainId, candidate.address, candidate.blockNumber)
+        fees = await extractFees__v3(candidate.chainId, candidate.address, strategies, candidate.blockNumber)
```
matheus1lva: I prefeered to go through fetching by "current" block number because im assuming fees can change, not too frequent but can. This is the major "bottleneck" as of right now. Unless there's already inside kong ways to get these per block.

murderteeth (changes requested):
### 1. Silent fee fetch failure in timeseries hook

`timeseries/apr-oracle/hook.ts:56-62` — When `projectStrategies` or `extractFees__v3` throws, fees silently default to `{ management: 0, performance: 0 }`, making `netApr === grossApr`. This writes incorrect data with no log. Add a `console.warn` so these are visible.

### 2. Backfill can use indexed fee-change events instead of per-row RPC

`backfill-apr-oracle-netapr.ts:125-131` — The current implementation correctly resolves fees at each candidate's block number, which handles fee changes over time. The bottleneck (as noted in your comment) is the RPC call per row — ~787k calls for chain 1.

Lets try this instead. Kong already indexes accountant fee-change events (`UpdateCustomFeeConfig`, `UpdateDefaultFeeConfig`) in `evmlog` via `packages/ingest/abis/yearn/3/accountant/abi.ts`. The script could query these events to build a fee timeline per vault, then look up the applicable fees for each candidate's `blockNumber` from the timeline — reducing RPC calls from ~787k to roughly the number of fee-change events (a handful per vault).

Please update with backfill performance numbers after this change.

murderteeth:
**Backfill procedure thought** — What do you think about running the backfill on a Neon branch db first, then upserting the results from the branch to prod in a second script? That way the backfill can run without touching prod, and the final upsert is just a bulk copy of pre-computed values — fast and low-risk.

matheus1lva:
> **Backfill procedure thought** — What do you think about running the backfill on a Neon branch db first, then upserting the results from the branch to prod in a second script? That way the backfill can run without touching prod, and the final upsert is just a bulk copy of pre-computed values — fast and low-risk.

That's a good point though.

matheus1lva:
> ### 1. Silent fee fetch failure in timeseries hook
> 
> `timeseries/apr-oracle/hook.ts:56-62` — When `projectStrategies` or `extractFees__v3` throws, fees silently default to `{ management: 0, performance: 0 }`, making `netApr === grossApr`. This writes incorrect data with no log. Add a `console.warn` so these are visible.
> 
> ### 2. Backfill can use indexed fee-change events instead of per-row RPC
> 
> `backfill-apr-oracle-netapr.ts:125-131` — The current implementation correctly resolves fees at each candidate's block number, which handles fee changes over time. The bottleneck (as noted in your comment) is the RPC call per row — ~787k calls for chain 1.
> 
> Lets try this instead. Kong already indexes accountant fee-change events (`UpdateCustomFeeConfig`, `UpdateDefaultFeeConfig`) in `evmlog` via `packages/ingest/abis/yearn/3/accountant/abi.ts`. The script could query these events to build a fee timeline per vault, then look up the applicable fees for each candidate's `blockNumber` from the timeline — reducing RPC calls from ~787k to roughly the number of fee-change events (a handful per vault).
> 
> Please update with backfill performance numbers after this change.

Ohhhhh so we have fee change, that's interesting. Yeah that can help pull to which periods had a "stable" fee. Hence reducing it. Thanks for the insight!

matheus1lva:
```
puts
upserted: 786900 timeseries outputs
bun --env-file=./.env run packages/scripts/src/backfill-apr-oracle-netapr.ts   5039.03s user 7331.28s system 100% cpu 3:24:25.83 total
```

murderteeth (changes requested):
### 1. Copy script drops `block_time` on write (High)

`copy-output-fork-to-main.ts` — The SELECT reads 8 columns including `block_time`, but the INSERT only writes 7 (omitting `block_time`). Copied rows will have NULL `block_time` in the main DB.

```
SELECT chain_id, address, label, component, value, block_number, block_time, series_time  ← reads block_time
INSERT INTO output (chain_id, address, label, component, value, block_number, series_time)  ← doesn't write it
```

### 2. Copy script pagination can infinite-loop (High)

`copy-output-fork-to-main.ts:63` — The cursor only advances by `block_time`, with `(0, '', '')` hardcoded for the remaining columns:

```sql
AND (block_time, chain_id, address, component) > ($3::timestamptz, 0, '', '')
```

Because `chain_id` is always > 0 and `address` is always > `''`, the next page re-reads almost every row with that same `block_time`. This causes duplicate rereads whenever multiple rows share the last batch's `block_time`, and an infinite loop if `READ_BATCH_SIZE` rows at that timestamp fill the next page again.

Fix: persist the full last tuple and compare against it:

```sql
AND (block_time, chain_id, address, component) > ($3::timestamptz, $4, $5, $6)
```

### 3. Backfill uses latest strategy set for all historical rows

`backfill-apr-oracle-netapr.ts:345-351` — `strategyCache` is populated once per vault at `maxBlock`. If a vault's strategy composition changed historically (strategies added/revoked), earlier rows will use the wrong strategy list for fee averaging via `resolveFees`. Consider building a strategy timeline from `StrategyChanged` events (same approach as the fee timelines) — the data is already in `evmlog`.

### 4. Backfill uses equal-weighted fee averaging instead of debt-weighted

`backfill-apr-oracle-netapr.ts:289-308` — `resolveFees` averages fees equally across strategies (`totalFee / strategies.length`), but the production logic in `extractFees__v3` (`apy.ts:349`) weights by `currentDebt / totalDebt`. Issue #363 calls for debt-ratio-weighted net values. The backfill will write wrong `netApr`/`netApy` for any vault with uneven debt distribution across strategies — even when strategy membership is unchanged.

murderteeth (approved):
lgtm

PR #368, opened 2026-03-17
Add busy-guard to ABI fanout to prevent queue bloat

### LINKED ISSUE #365: Add busy-guard to ABI fanout to prevent queue bloat
## Description

When the scheduled ABI fanout cron fires while extraction/load jobs from the previous cycle are still running, the fanout re-queues redundant work. While stride tracking prevents duplicate data writes, the queues bloat with redundant jobs that waste Redis memory and worker cycles. Add a guard that skips the fanout entirely when queues are still draining.

## Context

**Current implementation:** The ABI fanout is a BullMQ repeating job (default `*/15 * * * *`) configured in `config/abis.yaml`. When triggered, `packages/ingest/fanout/abis.ts` iterates all configured ABIs and queues `fanout.events`, `extract.snapshot`, and `fanout.timeseries` jobs for every source address. There is no check for whether the previous cycle's work has completed.

**Problem:** The `LatestBlocks` cron (`*/30s`) and `SystemProbe` cron (`*/10s`) always have jobs in the extract and probe queues respectively. A naive queue-depth check would cause perpetual skipping. The guard must use job-name-level filtering to distinguish background heartbeat jobs from real indexing work.

**Relevant references:**
- `packages/lib/mq.ts:24` — job name constants
- `packages/ingest/fanout/abis.ts` — fanout entry point
- `packages/ingest/fanout/index.ts` — fanout worker wiring
- `config/abis.yaml` — fanout cron schedule
- `config/crons.yaml` — LatestBlocks and SystemProbe crons
- `packages/lib/mq.ts:8` — existing Sentry init (scoped to MQ inventory metrics)

### 1. Add shared Sentry helper — `packages/lib/sentry.ts`

- Create a small module that initializes once from `SENTRY_DSN`
- Export a `captureMessage(message, options)` function
- No-op when Sentry is not configured
- Re-export from `packages/lib/index.ts`

### 2. Add busy-guard helper — `packages/ingest/fanout/isBusy.ts`

- Define type:
  ```ts
  type BusyMatch = { queue: string; jobName: string; status: 'waiting' | 'active' }
  ```

- Implement `findBusyMatch(): Promise<BusyMatch | null>` that scans queues using BullMQ `getJobs(['waiting', 'active'])` and bails on first qualifying match

- Implement `isBusy(): Promise<boolean>` as a thin wrapper over `findBusyMatch()`

- Queue scan rules (use `mq.job` constants for all name comparisons):

  | Queue | Include | Exclude |
  |---|---|---|
  | `fanout` | `events`, `timeseries` | `abis` |
  | `extract` (root) | `manuals`, `waveydb`, `webhook` | — |
  | `extract-{chainId}` (per-chain) | `evmlog`, `snapshot`, `timeseries` | `block` |
  | `load` | `evmlog`, `snapshot`, `thing`, `output`, `price`, `monitor` | `block` |
  | `probe` | — (skip entirely) | — |

- Only inspect `waiting` and `active` states. Ignore completed, failed, delayed, and repeatable definitions.

- Short-circuit: return on first qualifying job, don't scan remaining queues.

### 3. Wire guard into ABI fanout — `packages/ingest/fanout/abis.ts`

- Import `findBusyMatch` from `./isBusy`
- Import `captureMessage` from `lib/sentry`
- At the top of `fanout()` (line 6, before `extract.manuals` is enqueued):
  - Call `await findBusyMatch()`
  - If match found:
    - Log: `console.error('🚨 ABI_FANOUT_SKIPPED_BUSY: previous ingestion work is still active or queued, queue=${match.queue} job=${match.jobName} status=${match.status}')`
    - Send Sentry message at `warning` level with tags/extra:
      - `component: 'ingest'`
      - `job: 'fanout.abis'`
      - `reason: 'busy'`
      - `queue: match.queue`
      - `jobName: match.jobName`
      - `status: match.status`
    - Return immediately
  - Else: proceed with existing fanout logic unchanged

## Acceptance Criteria

- [ ] A slow `output`, `price`, `evmlog`, `snapshot`, `thing`, or `monitor` load job blocks the next ABI fanout
- [ ] `LatestBlocks` extract/load jobs (`block`) do not block ABI fanout
- [ ] `SystemProbe` jobs do not block ABI fanout
- [ ] The fanout `abis` job itself does not block (self-referential skip)
- [ ] Guard applies to both cron-driven and manual terminal-triggered fanout abis
- [ ] Console log is grep-friendly (`ABI_FANOUT_SKIPPED_BUSY`)
- [ ] Sentry receives a warning-level message (not exception) with queue/job/status details
- [ ] When Sentry is not configured, the guard still works (log-only, no crash)
- [ ] Existing fanout logic is unchanged when queues are idle

## Technical Notes

- Because exclusions are job-name-specific inside mixed queues, queue-level `getJobCounts()` is not sufficient. Must inspect individual jobs via `getJobs()`.
- BullMQ `getJobs` supports pagination — use it and bail on first match to avoid loading thousands of jobs into memory.
- The Sentry init in `packages/lib/mq.ts:10` is gated on `MQ_INVENTORY` and scoped to metrics. The new `lib/sentry.ts` helper should initialize independently from `SENTRY_DSN` alone.

matheus1lva:
### Summary

When the scheduled ABI fanout cron fires while extraction/load jobs from the previous cycle are still running, it re-queues redundant work. While stride tracking prevents duplicate writes, queues still bloat with redundant jobs and wasted worker cycles. This adds a guard that skips `fanout.abis` when qualifying ingest/load work is already `waiting` or `active`.

The guard uses job-name-level filtering so background heartbeat jobs do not trigger the skip. `fanout/abis`, `extract/block`, `load/block`, and `probe/all` are intentionally ignored.

Closes #365

### How to review

- `packages/ingest/fanout/abis.ts` — early return at the top of `fanout()`
- `packages/ingest/fanout/isBusy.ts` — queue scan and job-name allowlist
- `packages/lib/sentry.ts` — warning emission for skipped fanout runs

Review focus:
- only the intended queue/job combinations trigger the skip
- heartbeat/system jobs do not block fanout
- the skip path returns before enqueueing any follow-on work

### Test plan

- [x] Manual: busy queue blocks ABI fanout
  Validate against a live Kong stack, not smoke tests. Bring the stack up, wait for `/api/mq`, hold a real watched job such as `extract/manuals` in `active` state, then trigger `fanout.abis`. Expected:
  - ingest logs `ABI_FANOUT_SKIPPED_BUSY`
  - `/api/mq?queue=extract` still shows the existing busy job
  - `/api/mq?queue=fanout` remains empty (`active: 0`, `waiting: 0`)

- [ ] Manual: idle queues still allow fanout
  With no qualifying `waiting` or `active` jobs in watched queues, trigger `fanout.abis`. Expected:
  - no `ABI_FANOUT_SKIPPED_BUSY` log
  - follow-on work appears in `extract` and/or `fanout`

- [ ] Manual: ignored heartbeat jobs do not block fanout
  Leave only `extract/block`, `load/block`, or `probe/all` activity present, then trigger `fanout.abis`. Expected:
  - no `ABI_FANOUT_SKIPPED_BUSY` log
  - fanout proceeds normally

- [ ] Manual: Sentry integration is non-blocking
  Run the skip path once with `SENTRY_DSN` unset and once with it set. Expected:
  - no crash in either case
  - warning event emitted only when Sentry is configured

- [ ] Automated: not run for this PR update
  Validation was done with an isolated Kong stack plus queue/API inspection instead of smoke tests.

### Risk / impact

Main risk is over-matching and skipping ABI fanout when only benign background work exists. The review focus should be the allowlist in `isBusy.ts`.

Rollback is straightforward: remove the guard in `abis.ts` and the queue scan helper in `isBusy.ts`.


murderteeth (changes requested):
## Review — PR #368: Add busy-guard to ABI fanout

### Overall

The core busy-guard logic is solid — allowlist approach is correct, `findBusyMatch` short-circuits on first match, pagination prevents loading thousands of jobs, and the queue connection cleanup via `try/finally` is handled properly. The implementation matches the spec in #365 closely.

Two issues to address, one blocking and one minor.

---

### 1. Sentry side-effect on all `lib` consumers (blocking)

`packages/lib/sentry.ts` calls `Sentry.init()` at the top level when `SENTRY_DSN` is set. Since it's re-exported from `lib/index.ts`, **every process that imports from `lib`** (ingest, scripts, tests) will now load `@sentry/node` and potentially initialize it — even if that process has no reason to use Sentry.

This also creates a double-init conflict with the existing Sentry setup in `mq.ts:10-14`, which dynamically imports and inits Sentry gated on `MQ_INVENTORY`.

**Suggestion:** Make the init in `sentry.ts` lazy (only initialize on first use), and then refactor `mq.ts` to use the same `lib/sentry` helper instead of maintaining its own separate init path. That centralizes Sentry initialization in one place and eliminates the double-init risk.

### 2. Revert accidental commit `900ca7f`

The backfill script `packages/scripts/src/backfill-apr-oracle-netapr.ts` landed on this branch but is unrelated to the busy-guard feature and is already in another PR. Please revert `900ca7f` from this branch.

### What looks good

- **Allowlist design in `isBusy.ts`** — clean separation between queue specs and scan logic, uses `mq.job` constants consistently, correctly excludes `block`, `abis`, and `probe/all`
- **Short-circuit + pagination** — returns on first qualifying match, pages through jobs in batches of 100
- **Queue cleanup** — `try/finally` ensures `queue.close()` even on early return
- **`abis.ts` integration** — guard is at the very top of `fanout()`, before any work is enqueued
- **Log format** — `ABI_FANOUT_SKIPPED_BUSY` is grep-friendly with structured context
- **Sentry warning** (not exception) with proper tags/extra metadata

murderteeth (changes requested):
Commit `900ca7f` (`backfill-apr-oracle-netapr.ts`) is still on this branch. Please revert it per the previous review.

murderteeth (approved):

PR #374, opened 2026-03-19
fix: V2 vault fee extraction + filter bad report timestamps

0xeye:
## Summary

- **V2 vault fees**: The V2 snapshot hook now reads `managementFee()` and `performanceFee()` on-chain via multicall. Previously V2 vaults returned `fees: null` from the GraphQL API, causing downstream consumers (yearn-metrics) to have no fee configuration for ~300 V2 vaults. Typical V2 values: 0 mgmt / 2000 perf (0%/20%), verified against mainnet.

- **Report timestamp filter**: The `vaultReports` resolver now filters out ~3,022 rows where `block_time` contains block numbers instead of Unix timestamps (render as 1970 dates). All affected rows have `gainUsd=0` so no revenue data is lost. Added `AND block_time >= '2020-01-01'::timestamptz` to the WHERE clause.

## Test plan

- [ ] Verify V2 fee extraction: after next snapshot cycle, query `vaults(yearn: true)` and confirm V2 vaults now return `fees: { managementFee, performanceFee }` instead of `null`
- [ ] Verify report filtering: query `vaultReports(chainId: 1, address: "0xa258C4606Ca8206D8aA700cE2143D7db854D168c")` and confirm no results with blockTime before 2020
- [ ] Confirm no regression: V3 vault fees still populated correctly

### DIFF
--- a/packages/ingest/abis/yearn/2/vault/snapshot/hook.ts
+++ b/packages/ingest/abis/yearn/2/vault/snapshot/hook.ts
@@ -12,6 +12,25 @@ import { mq } from 'lib'
 import { compare } from 'compare-versions'
 import { getLatestApy, getLatestEstimatedApr } from '../../../../../helpers/apy-apr'
 
+async function extractV2Fees(chainId: number, address: `0x${string}`) {
+  try {
+    const abi = parseAbi([
+      'function managementFee() view returns (uint256)',
+      'function performanceFee() view returns (uint256)'
+    ])
+    const [mgmt, perf] = await rpcs.next(chainId).multicall({ contracts: [
+      { address, abi, functionName: 'managementFee' },
+      { address, abi, functionName: 'performanceFee' }
+    ] })
+    return {
+      managementFee: mgmt.status === 'success' ? Number(mgmt.result) : 0,
+      performanceFee: perf.status === 'success' ? Number(perf.result) : 0
+    }
+  } catch {
+    return { managementFee: 0, performanceFee: 0 }
+  }
+}
+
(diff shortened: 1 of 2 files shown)


murderteeth (changes requested):
## Review

### V2 fee extraction

V2 vaults already expose `managementFee` and `performanceFee` as top-level fields on the snapshot — it intentionally mirrors on-chain contract state:

```bash
curl -s 'https://kong.yearn.fi/api/rest/snapshot/1/0xa354F35829Ae975e850e23e9615b11Da1B3dC4DE' | jq '{managementFee, performanceFee, fees}'
```

These are actually already served through the API. The `fees` wrapper object is a V3 hook concern — V3 vaults need it because fees come from an external accountant contract, not the vault itself.

Kong is passthrough by default — it relays on-chain data as-is without normalizing across contract versions. This means no information is lost and there's no mapping layer that can drift out of sync with the contracts.

### Report timestamp filter

There's no evidence of bad `block_time` data. The earliest `StrategyReported` row is from 2021-02-01, well after the `2020-01-01` cutoff:

```bash
curl -s 'https://kong.yearn.fi/api/rest/reports/1/0xa258C4606Ca8206D8aA700cE2143D7db854D168c' | jq 'sort_by(.blockTime) | .[0].blockTime'
```

If you can provide specific examples of reports with bad `block_time` values we can revisit this.

### Note on test plans

Kong's REST API is the preferred access method and is generally faster than GraphQL.

---

With both changes addressed, I think we can close this one out. Thanks for digging into it!

PR #379, opened 2026-03-27
Add QA scripts for tvl gap detection and backfill

murderteeth:
## Summary
- `tvl-detect-gaps.ts` — scans vault timeseries for missing, zero, or incomplete tvl data points, classifies root cause (price, snapshot, computation), and reports by chain
- `tvl-backfill.ts` — recomputes tvl data for specific vaults using on-chain multicalls (totalAssets) and DefiLlama historical prices (tvls)
- README documenting options, examples, and typical detect → backfill → verify workflow


matheus1lva (approved):
Oh so the issue was not only timeseries which had wrong value or assets usd, but rather the price missing for a given vault. Understandable!

murderteeth:
> Oh so the issue was not only timeseries which had wrong value or assets usd, but rather the price missing for a given vault. Understandable!

yes! for some reason we weren't getting prices on those particular tokens on those particular days.

PR #385, opened 2026-04-08
Fix APY for vaults whose asset is a yield-bearing vault

matheus1lva:
### Summary

Vaults whose asset is itself another vault (e.g. Locked yvUSD wraps yvUSD) were only capturing the wrapper's PPS growth in the `apy-bwd-delta-pps` timeseries, missing the underlying vault's yield entirely.

For Locked yvUSD this meant showing ~4% APY instead of ~8.5% — the 4% was only the locking bonus, completely omitting the yvUSD base yield (~4.3%).

The fix composes PPS values when the vault's asset is a known Yearn vault (has `apiVersion`) in the `thing` table: `composedPPS = vaultPPS × assetPPS / 10^assetDecimals`.

### Affected vaults (10)

Inner vault must have `v3=true` set (excludes non-Yearn ERC4626 like sDAI which lacks `pricePerShare()`).

```
1	0x082a5743aAdf3d0Daf750EeF24652b36a68B1e9C
1	0x10B13EF16Cf838D0A290B5eecba11599DDa981D4
1	0x23346B04a7f55b8760E5860AA5A77383D63491cD
1	0x319C47a656BC2FCebB0cDDFF763A40195f70C157
1	0x7C7569585Ba5E7F5c0b1D18AB630D1769CA27193
1	0xAaaFEa48472f77563961Cdb53291DEDfB46F9040
1	0xCa36627878621FF63239E9819e357E41666317b6
1	0xb974598227660bEfe79a23DFC473D859602254aC
1	0xdCBa6f240Bb7e8Ea00F23a4F970B29d22DDd860E
137	0x4987d1856F93DFf29e08aa605A805FaF43dC3103
```

### How to review

**Core fix:** `packages/ingest/abis/yearn/lib/apy.ts`

1. The `_compute` function now checks if `vault.defaults.asset` is itself a vault in the `thing` table **with `apiVersion` set** (confirms it's a Yearn vault with `pricePerShare()`)
2. If so, at each block (current, inception, weekly, monthly), it reads `pricePerShare()` on the asset vault and multiplies into the composed PPS
3. The rest of the function (delta calculation, fee reversal for grossApr) works unchanged on the composed values

**Backfill scripts:** `packages/scripts/src/historical-apy-backfill/`

- `compute.ts` — reuses `_process` from `ingest/abis/yearn/lib/apy` (no reimplemented logic), writes to `output_temp_apy_backfill` staging table
- `upsert.ts` — promotes from staging to `output` and drops temp table in one transaction

### Test plan

- [ ] Start dev environment
```
make dev
```

- [ ] Trigger timeseries replay for Locked yvUSD to recompute APY with the fix
```
Ingest → fanout replays → Confirm
```
Expected: `apy-bwd-delta-pps` job runs for `0xAaaFEa48472f77563961Cdb53291DEDfB46F9040` without errors

- [ ] Verify composed PPS stored in output table
```sql
SELECT component, value
FROM output
WHERE chain_id = 1
  AND address = '0xAaaFEa48472f77563961Cdb53291DEDfB46F9040'
  AND label = 'apy-bwd-delta-pps'
  AND block_time = (
    SELECT MAX(block_time) FROM output
    WHERE chain_id = 1
      AND address = '0xAaaFEa48472f77563961Cdb53291DEDfB46F9040'
      AND label = 'apy-bwd-delta-pps'
  )
ORDER BY component;
```
Expected: `pricePerShare` ≈ 1046xxx (composed, not 1038xxx), `monthlyNet` ≈ 0.08 (not 0.04)

- [ ] Refresh snapshot cache and verify API response
```
```
```
curl -s http://localhost:3000/api/rest/snapshot/1/0xAaaFEa48472f77563961Cdb53291DEDfB46F9040 | jq '.apy'
```
Expected: `net` ≈ 0.08, `monthlyNet` ≈ 0.08, `pricePerShare` shows composed value

- [ ] Verify non-nested vaults are unaffected — check a regular vault
```
curl -s http://localhost:3000/api/rest/snapshot/1/0x696d02Db93291651ED510704c9b286841d506987 | jq '.apy'
```
Expected: `net` unchanged (~0.043), no composed PPS logic triggered (asset is USDC, not a vault)

### Post-deploy: historical APY recomputation

After the deploy lands on Render (`ingest-v-2` service), run the backfill scripts to recompute all historical APY data for the 22 affected vaults.

**0. Pause indexing**

**1. Open a shell on Render**

Go to [Render Dashboard](https://dashboard.render.com/) → `ingest-v-2` service → **Shell** tab.

**2. Dry run (preview)**
```
bun packages/scripts/src/historical-apy-backfill/compute.ts --vaults 1:0xAaaFEa48472f77563961Cdb53291DEDfB46F9040,1:0x722f76f34cB5c3B008A50E0664e55A53F4F461AB,1:0x23346B04a7f55b8760E5860AA5A77383D63491cD,1:0x63859212aa05d60295a2F18a9e0C707040605BAd,1:0x319C47a656BC2FCebB0cDDFF763A40195f70C157,1:0x082a5743aAdf3d0Daf750EeF24652b36a68B1e9C,1:0xb974598227660bEfe79a23DFC473D859602254aC,1:0x10B13EF16Cf838D0A290B5eecba11599DDa981D4,137:0x4987d1856F93DFf29e08aa605A805FaF43dC3103,1:0x093f4FCA4b71912EDb33B2d4Bb92E5b85658D833,1:0x1eB1d7fEc395748cFF0e437F4B3e8A850550987d,1:0x2BCc065433c608B2e71eA80405E260cA9B40ff65,1:0x45380B6249a1005bec67928AF03b8aA8ec4eadCC,1:0x4D840966887673132Ec61aB4Ffb5e3c26cB1F4BD,1:0x4E69d3b01a5693DBff8dC0AD340eBF816b38D162,1:0x66Afc16E2FaCf4EeA0712EDEBE0B96C5f51606Fa,1:0x7c37d1Ca716c4519BEcc9bC0CEea065C88caF98B,1:0x90B28aa7Df2eFaaAD4582afaF4D2a1c61519A245,1:0xED08B03bAeAB9EEF24F8AAa7c752f34a3264992e,1:0xb5907b7913F4dc10C5c0fc75c570c7D78D16768F,1:0xf4e7Ff66Fd3d1c33B3020fFA575C33Cb2968082D,1:0xf750c238cE033afC224C2E57d0fcb86E64Db0221 --dry-run
```
Expected: lists entry counts per vault, no DB writes.

**3. Compute (writes to staging table)**
```
bun packages/scripts/src/historical-apy-backfill/compute.ts --vaults 1:0xAaaFEa48472f77563961Cdb53291DEDfB46F9040,1:0x722f76f34cB5c3B008A50E0664e55A53F4F461AB,1:0x23346B04a7f55b8760E5860AA5A77383D63491cD,1:0x63859212aa05d60295a2F18a9e0C707040605BAd,1:0x319C47a656BC2FCebB0cDDFF763A40195f70C157,1:0x082a5743aAdf3d0Daf750EeF24652b36a68B1e9C,1:0xb974598227660bEfe79a23DFC473D859602254aC,1:0x10B13EF16Cf838D0A290B5eecba11599DDa981D4,137:0x4987d1856F93DFf29e08aa605A805FaF43dC3103,1:0x093f4FCA4b71912EDb33B2d4Bb92E5b85658D833,1:0x1eB1d7fEc395748cFF0e437F4B3e8A850550987d,1:0x2BCc065433c608B2e71eA80405E260cA9B40ff65,1:0x45380B6249a1005bec67928AF03b8aA8ec4eadCC,1:0x4D840966887673132Ec61aB4Ffb5e3c26cB1F4BD,1:0x4E69d3b01a5693DBff8dC0AD340eBF816b38D162,1:0x66Afc16E2FaCf4EeA0712EDEBE0B96C5f51606Fa,1:0x7c37d1Ca716c4519BEcc9bC0CEea065C88caF98B,1:0x90B28aa7Df2eFaaAD4582afaF4D2a1c61519A245,1:0xED08B03bAeAB9EEF24F8AAa7c752f34a3264992e,1:0xb5907b7913F4dc10C5c0fc75c570c7D78D16768F,1:0xf4e7Ff66Fd3d1c33B3020fFA575C33Cb2968082D,1:0xf750c238cE033afC224C2E57d0fcb86E64Db0221
```
Expected: recomputes APY using live `_process`/`_compute`, writes to `output_temp_apy_backfill` staging table.

**4. Verify staging table**
```
bun packages/scripts/src/historical-apy-backfill/upsert.ts --dry-run
```
Expected: shows row count, sample rows, distinct vault count. No changes made.

**5. Promote to output**
```
bun packages/scripts/src/historical-apy-backfill/upsert.ts
```
Expected: upserts all rows from staging into `output` and drops `output_temp_apy_backfill` in one transaction.

**6. Refresh snapshot cache**
```
```

**7. Verify**
**8. restart indexing**

```
curl -s https://kong.yearn.fi/api/rest/snapshot/1/0xAaaFEa48472f77563961Cdb53291DEDfB46F9040 | jq '.apy.net, .apy.monthlyNet'
```
Expected: both values ≈ 0.08 (not 0.04)

### Risk / impact

- **Low risk for non-nested vaults**: the `thing` table lookup returns `undefined` for standard vaults, so the entire composition block is skipped
- **Additional RPC calls**: up to 4 extra `pricePerShare` reads per nested vault per APY computation
- **Rollback**: revert the commit; stale APY values will remain until next recomputation


matheus1lva (commented):
## Summary
Fixes under-reported APY for 22 nested vaults (vaults whose asset is itself a Yearn vault) by composing outer × inner `pricePerShare` at each computation block. Includes a one-off backfill script to recompute historical `apy-bwd-delta-pps` outputs. Core math is sound and earlier review nits were cleanly addressed in 21a9ca3.

## Math verification
`composedPPS = vaultPps × assetPps / 10^innerDecimals` — units check out:
- Outer `pricePerShare()` → inner-vault units per outer-share (scaled to inner.decimals)
- Inner `pricePerShare()` → inner-asset units per inner-share (scaled to inner.decimals)
- Product / 10^inner.decimals → inner-asset units per outer-share

Since `compoundAndAnnualizeDelta` works on relative deltas, absolute scale doesn't matter for APY — only consistency across blocks, which holds. Magnitudes stay at ~1e18 so no precision loss round-tripping through Number in `_process`.

## Observations (non-blocking)

- [ ] **`packages/ingest/abis/yearn/lib/apy.ts:159` — address case sensitivity (latent).** The new query does a case-sensitive match:
  ```sql
  WHERE chain_id = $1 AND address = $2 AND label = $3 AND defaults->>'apiVersion' IS NOT NULL
  ```
  Meanwhile the backfill script uses `LOWER(t.defaults->>'asset') = LOWER(t2.address)` defensively. In practice both `thing.address` and `defaults.asset` come through the same viem-decoded + checksummed paths (via `zhexstring`/`EvmAddressSchema.transform(getAddress)`), so they should match today. But if any future code path writes a lowercased address into either field, this lookup silently returns `undefined` and composition gets skipped with no warning. Suggestion: either match the script's `LOWER()` approach for consistency, or add a `console.warn` when `vault.defaults.asset` is set but the lookup returns nothing, so silent misses become visible.

- [ ] **No unit test for composed PPS path.** `apy.spec.ts` has no `nested`/`compose`/`inner` coverage. Not blocking for a one-off fix, but a single spec mocking `rpcs.next()` and a nested `thing` row would protect against regressions.

## Suggestions

- **`packages/ingest/abis/yearn/lib/apy.ts:161`** — the early-return `if (result.pricePerShare === result.inceptionPricePerShare) return result` now compares composed values. Extremely edge-case (outer grows while inner shrinks such that the product equals inception exactly), harmless.

- **No try/catch around inner `pricePerShare` reads** — consistent with existing behavior for the outer reads, so not a regression. Worth knowing: if an inner-vault RPC read reverts at any block, the whole APY computation fails for that run.

- **grossApr fee claim** — PR body correctly notes outer fees apply to composed profit for strategy-based Yearn vaults. Worth a quick sanity check on the Cove Rewards Gauges (15 of 22 vaults) — if they're pass-through gauges with `fees = 0`, the math degenerates cleanly to `grossApr = netApr` anyway.

## Resolved in 21a9ca3

- `blocksPerDay` hardcoded `7200n` → now derives from weekly window, falls back to monthly-derived, otherwise skips the entry (fixes Polygon ~2s blocks)
- `grossApr` proportional scaling now documented as intentional approximation
- `viem` added as explicit dep in `scripts/package.json` (pinned to 2.5.0 per repo convention)

## Verdict
LGTM — core fix is sound, earlier review feedback fully addressed. Two observations above are hygiene improvements for a follow-up PR, not blockers.

---

## How This Was Reviewed

murderteeth (changes requested):
## Findings

**1. Lint error — blocking CI** (`packages/ingest/abis/yearn/lib/apy.ts:159`)

Template literal with no interpolation triggers the `quotes` rule. Needs single quotes with escapes or an inline disable.

**2. Wrong filter condition** (`packages/ingest/abis/yearn/lib/apy.ts:159`, `packages/scripts/src/historical-apy-yield-bearing/index.ts:163`)

`defaults->>'apiVersion' IS NOT NULL` should be `(defaults->'v3')::boolean IS TRUE`.

**3. Backfill script needs a rewrite** (`packages/scripts/src/historical-apy-yield-bearing/`)

Three problems:

- **Duplicated logic.** The script reimplements the APY computation instead of reusing `_compute` from `packages/ingest/abis/yearn/lib/apy.ts`. See `packages/scripts/src/backfill-apr-oracle-netapr.ts` for the established pattern — it imports from `ingest/abis/yearn/lib/apy`, `ingest/db`, and `ingest/load` directly.

- **Wrong interface.** The script auto-discovers nested vaults and hardcodes its scope. It should follow the same pattern as `packages/scripts/src/quality-assurance/tvl-backfill.ts`: take an explicit vault list via `--vaults chainId:0xABC,...`, a `--start`/`--end` date range, and `--dry-run`. This should be a general-purpose APY backfill tool, not specific to nested vaults.

- **No verification step.** The script upserts directly into the `output` table. Split into two scripts under `historical-apy-backfill/`: `compute.ts` recomputes APY and loads results into an `output_temp` table, `upsert.ts` bulk upserts from `output_temp` into `output` and drops `output_temp` in the same transaction.

matheus1lva:
@murderteeth All three findings addressed:

**1. Lint error** — Fixed in 210f977. Converted the template literal to a single-quoted string (no interpolation was needed).

**2. Wrong filter condition** — Already fixed in 76a2297. The query uses `(defaults->'v3')::boolean IS TRUE` (not `defaults->>'apiVersion' IS NOT NULL`).

**3. Backfill script rewrite** — Already done in 5d5cb13. The script at `packages/scripts/src/historical-apy-backfill/` now:
- **Reuses `_process`** from `ingest/abis/yearn/lib/apy` (no duplicated logic)
- **Takes explicit CLI args**: `--vaults chainId:0xABC,...`, `--start`, `--end`, `--dry-run`
- **Splits into two scripts**: `compute.ts` writes to `output_temp_apy_backfill`, `upsert.ts` promotes to `output` and drops the temp table in a single transaction

murderteeth (approved):
Before merge:

- [x] Post an updated list of affected vaults — I can see many of the vaults in the post-merge script have v3=null, e.g. `0x4D840966887673132Ec61aB4Ffb5e3c26cB1F4BD`
- [x] Verify we want Staked yBOLD (`0x23346B04a7f55b8760E5860AA5A77383D63491cD`) to use new apy logic
- [ ] Update the post-deploy plan to pause the indexer before running the backfill recomputation and unpause after — we don't want historical APY to look out of sync on yvUSD or any others updated by the script

murderteeth:
please update the PR notes to show both the address and name of affected vaults like it did before. also make sure the scripts are updated, they still show the 22 vault addresses from before  @matheus1lva

matheus1lva:
> please update the PR notes to show both the address and name of affected vaults like it did before. also make sure the scripts are updated, they still show the 22 vault addresses from before @matheus1lva

Yep, will add a follow up pr with some smol fixes and concurrency i used with that list updated plus all trail from this pr to that one

matheus1lva:
### Updated affected vault list

After running the discovery query against prod, the actual affected vault count is **10** (not 22). The original list included vaults whose outer vault is not v3, and 3 vaults whose inner vault lacks `v3=true`.

**Discovery query:**
```sql
SELECT t.chain_id, t.address
FROM thing t
JOIN thing t2
  ON t2.chain_id = t.chain_id
  AND LOWER(t2.address) = LOWER(t.defaults->>'asset')
  AND t2.label = 'vault'
WHERE t.label = 'vault'
  AND (t.defaults->'v3')::boolean IS TRUE
ORDER BY t.chain_id, t.address;
```

**Affected vaults (10):**

| Chain | Outer Vault |
|---|---|
| 1 | `0xb974598227660bEfe79a23DFC473D859602254aC` |
| 1 | `0xAaaFEa48472f77563961Cdb53291DEDfB46F9040` |
| 1 | `0xdCBa6f240Bb7e8Ea00F23a4F970B29d22DDd860E` |
| 1 | `0xCa36627878621FF63239E9819e357E41666317b6` |
| 1 | `0x7C7569585Ba5E7F5c0b1D18AB630D1769CA27193` |
| 1 | `0x319C47a656BC2FCebB0cDDFF763A40195f70C157` |
| 1 | `0x10B13EF16Cf838D0A290B5eecba11599DDa981D4` |
| 1 | `0x23346B04a7f55b8760E5860AA5A77383D63491cD` |
| 1 | `0x082a5743aAdf3d0Daf750EeF24652b36a68B1e9C` |
| 137 | `0x4987d1856F93DFf29e08aa605A805FaF43dC3103` |

**Also fixed:** address case sensitivity in the inner vault lookup — `getAddress()` now checksums `vault.defaults.asset` before querying, so all matching vaults are found regardless of stored casing.

### Backfill scripts

The backfill scripts had several runtime bugs (missing `rpcs.up()`, wrong timestamp types, duplicate key errors). These are fixed in #387.

PR #390, opened 2026-04-12
fix: add getCurrentApr fallback for vaults where getStrategyApr reverts

matheus1lva:
### Summary

The apr-oracle timeseries hook originally only called `getStrategyApr(vaultAddress, 0)` on the APR oracle contract. For vaults that aren't tokenized strategies (e.g. DAI-1 yVault `0x028eC7330ff87667b6dfb0D94b954c820195336c`), no strategy oracle is registered so `getStrategyApr` **reverts**. The catch block silently set `apr = 0`, producing faulty zero rows in the output table.

The fix keeps `getStrategyApr` as the primary call (it works for tokenized strategies) and adds a `getCurrentApr` fallback — the oracle's vault-level function that returns the APR based on the current profit-unlocking rate. This covers regular vaults where `getStrategyApr` reverts.

Includes three-phase backfill scripts (probe → compute → upsert) to fix existing output records.

### Root cause

```
Old hook: getStrategyApr(vaultAddress, 0) → REVERTS for non-strategy vaults → catch { apr = 0 }
Fix:      getStrategyApr(vaultAddress, 0) → direct strategy APR for tokenized strategies
          ↳ fallback: getCurrentApr(vaultAddress) → profit-unlocking rate APR for regular vaults
```

### How to review

1. `packages/ingest/abis/yearn/3/vault/timeseries/apr-oracle/hook.ts` — the core fix: `getStrategyApr` first, `getCurrentApr` fallback for vaults where `getStrategyApr` reverts. `readApr()` is exported so backfill scripts share the same logic.
2. `packages/scripts/src/backfill-apr-oracle-getCurrentApr/probe.ts` — diagnostic: calls `getStrategyApr` at the latest block for each vault with snapshot `oracle.apr=0`, identifies faulty vaults (those that revert), writes them to `probe-results.json`
3. `packages/scripts/src/backfill-apr-oracle-getCurrentApr/compute.ts` — re-queries oracle with the fixed `readApr()` logic at each historical block for faulty vaults, writes to temp table (supports `--from-probe` and pause/resume)
4. `packages/scripts/src/backfill-apr-oracle-getCurrentApr/upsert.ts` — promotes temp table to production output table

### Affected vaults query

The probe identifies faulty vaults from the `snapshot` table (v3 + yearn vaults with `oracle.apr = 0`):

```sql
SELECT s.chain_id, s.address
FROM snapshot s
JOIN thing t ON s.chain_id = t.chain_id AND s.address = t.address
WHERE s.hook #>> '{performance,oracle,apr}' IS NOT NULL
  AND (s.hook #>> '{performance,oracle,apr}')::numeric = 0
  AND t.label = 'vault'
  AND COALESCE((t.defaults->>'v3')::boolean, false)
  AND COALESCE((t.defaults->>'yearn')::boolean, false);
```

### Test plan

- [x] create `config/abis.local.yaml` to target the affected vault
```yaml
cron:
  name: AbiFanout
  queue: fanout
  job: abis
  schedule: '*/15 * * * *'
  start: false

abis:
  - abiPath: 'yearn/3/vault'
    sources: [
      { chainId: 1, address: '0x028eC7330ff87667b6dfb0D94b954c820195336c', inceptBlock: 19419991 }
    ]
```

- [x] create `config/chains.local.yaml` to restrict to mainnet
```yaml
chains: [
  mainnet
]
```

- [x] create `config/manuals.local.yaml` to restrict to mainnet
```yaml
manuals:
  - chainId: 1
    address: '0x028eC7330ff87667b6dfb0D94b954c820195336c'
    label: 'vault'
    defaults: {
      erc4626: true,
      v3: true,
      yearn: true,
      inceptBlock: 19419991,
    }

```

- [x] start full dev environment
```
make dev
```

- [x] trigger fanout to run the apr-oracle timeseries hook with the fix
```
```
Expected: jobs dispatched without errors, extract/load jobs process the vault

- [x] run the snapshot refresh script to populate the REST cache with updated snapshot data
```
```
Expected: snapshot cache refreshed with oracle data from the hook

- [x] verify oracle data is now populated in the snapshot
```
curl -s http://localhost:3001/api/rest/snapshot/1/0x028eC7330ff87667b6dfb0D94b954c820195336c | jq '.performance.oracle'
```
Expected: `apr` and `apy` are non-zero values (profit-unlocking rate APR from oracle), e.g.:
```json
{
  "apr": 0.03,
  "apy": 0.03,
  "netAPR": 0.027,
  "netAPY": 0.027
}
```

#### Backfill script
- [ ] point your local .env to a fork of the database

- [ ] run compute
```
bun packages/scripts/src/backfill-apr-oracle-getCurrentApr/compute.ts
```
Expected: computes corrected APRs, writes to `output_temp_apr_oracle_backfill` table

- [ ] run upsert 
```
bun packages/scripts/src/backfill-apr-oracle-getCurrentApr/upsert.ts
```

### Post-merge actions

1. **Deploy** the hook fix (merge this PR)

2. **Run compute**:
   ```
   bun packages/scripts/src/backfill-apr-oracle-getCurrentApr/compute.ts
   ```

5. **Apply upsert** (promotes temp table → output, drops temp table):
   ```
   bun packages/scripts/src/backfill-apr-oracle-getCurrentApr/upsert.ts
   ```

6. **Trigger snapshot refresh** so updated oracle values propagate to the REST cache

7. New timeseries runs will use `getStrategyApr` → `getCurrentApr` fallback automatically going forward


murderteeth (changes requested):
## Summary
Fixes a silent-zero bug in the v3 vault apr-oracle timeseries hook: it now prefers `getCurrentApr(vault)` (weighted average across strategies) and falls back to `getStrategyApr(address, 0)` on revert for tokenized strategies. Also adds a three-phase backfill (`probe` → `compute` → `upsert`) that writes corrected values into a temp table and promotes them to `output`.

## Issues
- [ ] **High — hook vs compute fallback divergence.** `packages/ingest/abis/yearn/3/vault/timeseries/apr-oracle/hook.ts:22-46` only falls back to `getStrategyApr` when `getCurrentApr` *throws*. `packages/scripts/src/backfill-apr-oracle-getCurrentApr/compute.ts:135-182` also falls back when `getCurrentApr` returns `0`. If the unsupported/tokenized-strategy vaults fail via a silent zero instead of a revert, the backfill will repair history but live indexing will keep writing zeros after deploy. Align the two: either add zero-fallback to the hook, or drop it from `compute.ts` with a rationale.

- [ ] **High — probe doesn't detect the documented failure mode.** `packages/scripts/src/backfill-apr-oracle-getCurrentApr/probe.ts:88-112` only reruns `getStrategyApr` and marks a vault faulty when that call reverts (`probe.ts:188-194`). But the PR describes the bug as a *silent zero return*, and `packages/scripts/src/backfill-apr-oracle-getCurrentApr/README.md:13` promises the probe checks `getCurrentApr` / `getStrategyApr` and identifies vaults with non-zero on-chain APR. As written, `compute.ts --from-probe` can miss affected vaults and leave bad zero rows unbackfilled. The probe should actually call `getCurrentApr` and flag vaults where it returns non-zero while stored rows are zero.

- [ ] **Test plan doesn't reproduce the fix.** Running the PR's verification step:
  ```
  $ curl -s http://localhost:3001/api/rest/snapshot/1/0x028eC7330ff87667b6dfb0D94b954c820195336c | jq '.performance.oracle'
  parse error: Invalid numeric literal at line 1, column 4
  $ curl -s http://localhost:3001/api/rest/snapshot/1/0x028eC7330ff87667b6dfb0D94b954c820195336c
  Not found
  ```
  The snapshot endpoint returns `Not found` for the affected vault after following the local setup (`config/abis.local.yaml` + `config/chains.local.yaml` + `make dev` + fanout + replay). The test plan needs to include whatever step is missing (e.g., a `thing` registration, an initial snapshot run, or correct local config wiring), otherwise reviewers can't actually verify the fix end-to-end.

- [ ] **packages/ingest/abis/yearn/3/vault/timeseries/apr-oracle/hook.ts:34** — Comment is indented 4 spaces; surrounding block is 6. Cosmetic.

- [ ] **PR title** — still prefixed with `DRAFT:` but the PR is not marked draft. Remove before merge.

## Verdict
REQUEST_CHANGES

---

## How This Was Reviewed

murderteeth (changes requested):
## Summary
Fixes apr-oracle timeseries hook silently storing `0` APR for v3 vaults that aren't tokenized strategies. When `getStrategyApr(vault, 0)` reverts, the new code falls back to `getCurrentApr(vault)`. Ships a three-phase backfill (probe → compute → upsert) to correct existing `output` rows.

## Issues

- [ ] **packages/scripts/src/backfill-apr-oracle-getCurrentApr/probe.ts:147** — TypeScript break in new code. `availableChainIds` is inferred from `chains.map(c => c.id)` as `Set<1 | 10 | 137 | 146 | 250 | 100 | 8453 | 42161 | 80094 | 747474>` (viem chain ids are literal unions), while `v.chainId` is a plain `number`. `availableChainIds.has(v.chainId)` fails strict type-check:
  ```
  probe.ts(147,32): error TS2345: Argument of type 'number' is not assignable
    to parameter of type '1 | 146 | 80094 | 747474 | 100 | 42161 | 8453 | 250 | 10 | 137'.
  ```
  Reproduce with `bunx tsc --noEmit` from `packages/scripts/`. `packages/scripts` isn't covered by CI lint/typecheck, so this merges green but leaves the new script broken. Fix: type the Set explicitly, e.g. `new Set<number>(chains.map(c => c.id))`.

- [ ] **packages/ingest/abis/yearn/3/vault/timeseries/apr-oracle/hook.ts:34** and **PR description** — the fallback comment and PR body describe `getCurrentApr` as "weighted average APR across all strategies," but per the canonical contract (`tokenized-strategy-periphery/src/AprOracle/AprOracle.sol:134-138`), `getCurrentApr(v) = getExpectedApr(v, 0)` — i.e., APR derived from the vault's **profit-unlocking rate** (already-reported profit still unlocking). The weighted-sum-of-strategy-APRs value comes from a different function, `getWeightedAverageApr(v, 0)` (AprOracle.sol:189). If the intent is the weighted average of strategy APRs, this is calling the wrong function; if the intent is the profit-unlocking APR, the comment and PR description are misleading. Please clarify which semantics you want and align the code/docs.

- [ ] **packages/scripts/src/backfill-apr-oracle-getCurrentApr/probe.ts:16-17** — docstring says "Uses the same `getCurrentApr → getStrategyApr` fallback logic as the apr-oracle hook." This is wrong two ways: (1) the probe intentionally does *not* fall back — it only calls `getStrategyApr` so it can detect the revert; (2) the hook's actual order is `getStrategyApr → getCurrentApr`, not the reverse.

- [ ] **PR description — Test plan / Post-merge actions look pasted wrong.** The "Test plan" and "Post-merge actions" sections have literal `\`` escapes around every inline-code and code-fence (e.g. ``create \`config/abis.local.yaml\``` and ` \`\`\`yaml `) instead of real backticks. On GitHub those render as raw `\` characters rather than code blocks, making the reproduction steps hard to follow. Please re-paste those sections with unescaped backticks.

## Suggestions

- `readApr` is duplicated verbatim between `hook.ts:14` and `compute.ts:135`. Export it from the hook and import in `compute.ts` so the fallback logic can't drift (also addresses the hook-vs-compute divergence from the prior review).

## Verdict
REQUEST_CHANGES

---

## How This Was Reviewed
This review was conducted using the [review-pr skill](https://github.com/yearn/webops-skills/blob/main/skills/review-pr/SKILL.md) and the [yearn-vaults skill](https://github.com/yearn/webops-skills/blob/main/skills/yearn-vaults/SKILL.md).

murderteeth (changes requested):
## Summary
Prior review at f39d964 flagged 4 items + 1 suggestion. As of f2dc626 only the suggestion (dedup `readApr`) was addressed. The 4 items are untouched, and no reply acknowledged them on the thread.

## Unaddressed from the previous review

- [ ] **packages/scripts/src/backfill-apr-oracle-getCurrentApr/probe.ts:147** — TypeScript still broken. `bunx tsc --noEmit` in `packages/scripts/`:
  ```
  probe.ts(147,32): error TS2345: Argument of type 'number' is not
    assignable to parameter of type '1 | 146 | 80094 | 747474 | 100 |
    42161 | 8453 | 250 | 10 | 137'.
  ```
  `probe.ts:56` is still `new Set(chains.map(c => c.id))`. The previous review gave the exact fix: `new Set<number>(chains.map(c => c.id))`. Unchanged since f39d964.

- [ ] **packages/ingest/abis/yearn/3/vault/timeseries/apr-oracle/hook.ts:34 and PR description** — `getCurrentApr` still described as "weighted average APR across all strategies." Per `tokenized-strategy-periphery/src/AprOracle/AprOracle.sol:134-138`, `getCurrentApr(v) = getExpectedApr(v, 0)` — the profit-unlocking rate. The weighted-average-of-strategy-APRs value is `getWeightedAverageApr(v, 0)` at `AprOracle.sol:189`. Either the code is calling the wrong function or the comment and PR body are wrong. The previous review asked for clarification + alignment. Neither happened.

- [ ] **packages/scripts/src/backfill-apr-oracle-getCurrentApr/probe.ts:16-17** — Docstring still claims "Uses the same `getCurrentApr → getStrategyApr` fallback logic as the apr-oracle hook." Still wrong two ways: the probe does not fall back (only calls `getStrategyApr` to detect the revert), and the hook's real order is `getStrategyApr → getCurrentApr`. Unchanged since the previous review pointed this out.

- [ ] **PR description — Test plan.** Formatting is still wrong, and the plan is missing steps. Following it end-to-end does not reproduce the fix. Please walk through it on a clean checkout and correct it.

## Process

Pushing a new commit without responding to a REQUEST_CHANGES review leaves the thread ambiguous — reviewers can't tell whether an item was considered and declined, forgotten, or disagreed with. For each open item, either push a fix, reply with why it's wrong, or reply that you're deferring it. Silent re-requests for review burn reviewer time because we have to re-verify the whole diff against the prior review to find out nothing changed.

## Verdict
REQUEST_CHANGES

---

## How This Was Reviewed

murderteeth (changes requested):
## Issues

- [ ] **packages/ingest/abis/yearn/3/vault/timeseries/apr-oracle/hook.ts:32-58 — wrong primary call.** The default entry point on Yearn's v3 AprOracle is `getStrategyApr(target, debtChange)`, not `getCurrentApr`. Per `tokenized-strategy-periphery/src/AprOracle/AprOracle.sol:134-138`, `getCurrentApr(v)` is literally `getExpectedApr(v, 0)` — APR implied by already-reported profit currently unlocking. `getStrategyApr` (`AprOracle.sol:61-101`) is the dispatch function: it consults per-strategy custom oracles (including the legacy registry), and only cascades into `getWeightedAverageApr`/`getExpectedApr` as a last resort. Calling `getCurrentApr` first skips that dispatch and writes the less-informative "reported-profit-unlocking" figure whenever the call succeeds. The prior hook's mistake was silently swallowing `getStrategyApr`'s revert and writing `apr=0`; the correct fix is to keep `getStrategyApr(address, 0n)` primary and handle revert by either falling back to `getCurrentApr` or returning `[]`. The oracle naming is unfortunate — please add a comment documenting why `getStrategyApr` is the default to prevent a future regression.

- [ ] **packages/scripts/src/backfill-apr-oracle-getCurrentApr/probe.ts:135 + README.md:57 — unsafe probe-gated workflow.** `probe.ts` classifies a vault as "faulty" only if `readApr()` returns non-zero at the *latest* block, and `compute.ts --from-probe` then filters to only those vaults (`compute.ts:259`). But `getCurrentApr` / `getExpectedApr` explicitly return 0 when profit unlocking has finished (`AprOracle.sol:162`). A vault between reports therefore reads `apr=0` at the latest block, gets classified as "genuine zero," and is dropped — even though it may have many historical rows incorrectly zeroed during periods when profit was actively unlocking. The probe-gating silently leaves those historical rows un-backfilled.

- [ ] **Test plan still does not reproduce the fix** (re-raising — flagged in prior reviews at f39d964 and f2dc626, unaddressed). Running the PR's step-7 verification:
  ```
  curl -s http://localhost:3001/api/rest/snapshot/1/0x028eC7330ff87667b6dfb0D94b954c820195336c | jq '.performance.oracle'
  parse error: Invalid numeric literal at line 1, column 4
  ```
  The snapshot endpoint still returns non-JSON (likely `Not found`) after following the local setup exactly as documented. Reviewers cannot verify the fix end-to-end. Walk through the plan on a clean checkout, identify the missing step, correct the plan, and mark each step done as you verify.

- [ ] **packages/scripts/src/backfill-apr-oracle-getCurrentApr/probe.ts:95 — likely case-mismatch in `getZeroRowCount`.** `WHERE address = $2` is queried with `address.toLowerCase()`, but live ingest writes through `OutputSchema` → `EvmAddressSchema.transform(s => getAddress(s))` (`packages/lib/types.ts:358`), producing checksummed strings; the column is plain `text` (case-sensitive). If apr-oracle rows are checksummed in prod (as the ingest path implies), `getZeroRowCount` returns 0 for every vault and the `zeroRows=…` / `total: N output rows to backfill` line in probe output is misleading. Confirm with a quick `SELECT address FROM output WHERE label='apr-oracle' AND component='apr' LIMIT 5` on prod, or drop the `.toLowerCase()` defensively. (Lower priority — becomes moot if probe-gating is removed per the item above.)

## Suggestions

- **Consider dropping the probe/filter step entirely and recomputing every v3 apr-oracle series.** The probe + `--from-probe` gating tries to be clever about "which vaults need fixing," but as shown above, the gate is unsafe (latest-block sample misses vaults between reports). A simpler and strictly-safer flow: recompute every v3 vault's `apr-oracle` output rows at their original block numbers via the corrected `readApr`, upsert the full result set, and let the ON CONFLICT update handle the overwrite. Correct values overwrite correct values idempotently; broken zeros get repaired regardless of latest-block state. This also collapses three scripts into one, removes the `probe-results.json` artifact, and eliminates a class of "did operators run probe recently enough?" questions. 

## Verdict
REQUEST_CHANGES

---

## How This Was Reviewed
This review was conducted using the [review-pr skill](https://github.com/yearn/webops-skills/blob/main/skills/review-pr/SKILL.md) and the [yearn-vaults skill](https://github.com/yearn/webops-skills/blob/main/skills/yearn-vaults/SKILL.md).

matheus1lva:
## Review findings + fixes applied

Reviewed the current state of this PR (post-41f4c3a). Two issues found and fixed in c5b15ac:

### 1. Misleading `getCurrentApr` comment (hook.ts:47-48)

The comment said "multi-strategy vaults" but the actual reason `getStrategyApr` reverts is that no strategy oracle is registered for the address (regular vault, not a tokenized strategy). Per `AprOracle.sol`, `getCurrentApr(v) = getExpectedApr(v, 0)` — returns APR based on the vault's profit-unlocking rate, not a weighted average of strategy APRs.

**Fixed:** Updated the comment to accurately describe the fallback mechanism.

### 2. Backfill README — probe-gated workflow gap

The README only documented the `probe → compute --from-probe → upsert` workflow. The `--from-probe` path only catches vaults where `getStrategyApr` currently reverts. A vault that historically reverted but since had a strategy oracle registered would be classified as "genuine" by the probe and skipped, leaving historical zero rows un-backfilled.

**Fixed:** Added a "Full recompute" section as the recommended default (runs `compute.ts` without `--from-probe` to process all zero rows). Kept the probe-gated workflow as a faster alternative with a note about its limitations.

### Note: test plan reproducibility

Prior reviews flagged that the test plan's snapshot endpoint returns `Not found` for the example vault after following the documented setup. This still needs to be walked through on a clean checkout to identify missing steps (thing registration, initial snapshot, or config wiring).

murderteeth (approved):
## Comment

- [ ] **PR description is stale.** The body still references a three-phase probe→compute→upsert workflow and links to `probe.ts` in "How to review" (item 2) and "Post-merge actions" (steps 2-3, including `--from-probe` flag). probe.ts was removed and compute.ts doesn't accept `--from-probe`. The README correctly describes the current two-phase flow — the PR body should match to avoid confusing operators following post-merge instructions.

## Verdict
APPROVE

PR #393, opened 2026-04-20
ci: add retry job to refresh cron workflows

matheus1lva:
## Summary
- Adds a `refresh_retry` job to each refresh workflow (lists, snapshot, timeseries, reports, and both historical variants) that runs only when the primary `refresh` job fails. This absorbs transient GitHub infra failures like action-download errors (e.g. `Failed to download archive '.../oven-sh/setup-bun/...'`).
- Consolidates Uptime Kuma reporting into a final `notify` job so success/failure is based on whether *either* attempt succeeded, instead of firing from the primary job that just failed.
- Historical workflows get the retry job but no notify job, since they didn't have Uptime Kuma reporting before.

## Test plan

- [ ] Trigger `refresh-lists` manually and confirm only the `refresh` job runs on success
\`\`\`
gh workflow run refresh-lists.yml
gh run list --workflow=refresh-lists.yml --limit 1
\`\`\`
Expected: single run with `refresh` + `notify` jobs, no retry job executed.

- [ ] Simulate failure (temporarily break the refresh command on a test branch) and confirm `refresh_retry` runs
\`\`\`
gh workflow run refresh-lists.yml --ref <test-branch>
gh run view <run-id>
\`\`\`
Expected: `refresh` fails, `refresh_retry` runs, `notify` reports based on retry result.

- [ ] Verify Uptime Kuma gets exactly one status update per workflow run (no duplicates from both attempts reporting)

### DIFF
--- a/.github/workflows/refresh-lists.yml
+++ b/.github/workflows/refresh-lists.yml
@@ -25,11 +25,38 @@ jobs:
           POSTGRES_POOL_MAX: ${{ vars.POSTGRES_POOL_MAX }}
           REST_CACHE_REDIS_URL: ${{ secrets.REST_CACHE_REDIS_URL }}
         run: bun packages/web/app/api/rest/list/refresh.ts
+
+  refresh_retry:
+    needs: refresh
+    if: failure()
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: oven-sh/setup-bun@v1
+      - name: Install dependencies
+        run: bun install --frozen-lockfile
+      - name: Refresh lists cache
+        env:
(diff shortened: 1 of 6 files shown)


murderteeth (changes requested):
- why are we using retries? what evidence motivated this
- the original issue suggests it may be a timing issue. is that theory wrong?
- why aren't historical workflows monitored?

matheus1lva:
At the time of the report it was a random github failure, it was exitting during setup phase, not on any timeout or anything similar - But as i looked again, the reports are failing and apparently that is a timing thing. 

The initial intent remains, there's nothing to do because of github failing - now im gonna check the other ones.

PR #394, opened 2026-04-22
fix: floor net APR at half gross APR

matheus1lva:
### Summary

Floor `computeNetApr` at `grossApr / 2` to match Yearn's accountant invariant (management + performance fees cap at 50% of profit), and return `0` for non-positive gross (the `grossApr / 2` floor can't meaningfully be negative). Previously the hook returned `(gross - mgmt) * (1 - perf)` unconditionally, which went negative whenever the management fee (charged on TVL) exceeded a very small gross APR. Historical rows stored with below-floor `netApr` / `netApy` are rewritten by a new backfill script.

### How to review

- Core change: `packages/ingest/abis/yearn/lib/apy.ts` — `computeNetApr` now guards non-positive gross with `return 0` and floors otherwise at `grossApr / 2`.
- Unit tests: `packages/ingest/abis/yearn/lib/apy.spec.ts` — existing `100% performance fee` case updated (now returns `gross / 2`, not `0`), plus new cases for the real-world `0xAa0362…` / katana inputs and for negative gross.
- Historical cleanup: `packages/scripts/src/backfill-apr-oracle-negative-netapr/` — `compute.ts` stages rewrites to a temp table (only `netApr` / `netApy` rows; companion `apr` / `apy` rows are left untouched), `upsert.ts` promotes them and drops the temp.
- Shared helper: `packages/scripts/src/backfill-shared/upsert.ts` centralizes the preview + upsert + drop flow. The sibling `backfill-apr-oracle-getCurrentApr/upsert.ts` was migrated onto it.

Production scope of the backfill (from Neon):

Around 653 addresses spread across `mainnet`, `katana`, `arbitrum` and 80094 

### Test plan

- [x] e2e on a real vault — paste the following into the three local YAML files (all gitignored), then `make dev`, and in the terminal UI: `Ingest → fanout abis → yes` and `Ingest → fanout replays → yes`.

`config/chains.local.yaml`
```yaml
chains: [
  'mainnet',
  'katana',
]
```

`config/abis.local.yaml`
```yaml
cron:
  name: AbiFanout
  queue: fanout
  job: abis
  schedule: '*/15 * * * *'
  start: false

abis:
  - abiPath: 'yearn/3/vault'
    sources: [
      { chainId: 747474, address: '0xAa0362eCC584B985056E47812931270b99C91f9d', inceptBlock: 4222239 }
    ]
```

`config/manuals.local.yaml`
```yaml
manuals:
  - chainId: 747474
    address: '0xAa0362eCC584B985056E47812931270b99C91f9d'
    label: 'vault'
    defaults: {
      erc4626: true,
      v3: true,
      yearn: true,
      inceptBlock: 4222239,
    }
```

- [x] verify no negatives in the katana vault's timeseries
```
PGPASSWORD=password psql -h localhost -p 5432 -U user -d user -c "SELECT component, count(*), min(value)::float, max(value)::float FROM output WHERE chain_id=747474 AND lower(address)=lower('0xAa0362eCC584B985056E47812931270b99C91f9d') AND label='apr-oracle' GROUP BY component ORDER BY component; "
```
Expected: `netApr` and `netApy` rows all `≥ 0` (in my run: 31 rows each, all equal to `gross / 2` ≈ `1.4e-6 … 5.6e-4`).

- [x] backfill — compute (stages rewrites, prod DB creds in `.env`)
```
bun packages/scripts/src/backfill-apr-oracle-negative-netapr/compute.ts
```

- [ ] backfill — apply
```
bun packages/scripts/src/backfill-apr-oracle-negative-netapr/upsert.ts
```
Expected: `found 184 rows in output_temp_netapr_floor_backfill`, `distinct vaults: 4`, `upserted 184 rows`, `dropped output_temp_netapr_floor_backfill`, `done.`

- [ ] verify no negatives remain in prod
```
PGPASSWORD="$POSTGRES_PASSWORD" PGSSLMODE=require psql -h "$POSTGRES_HOST" -p "$POSTGRES_PORT" -U "$POSTGRES_USER" -d "$POSTGRES_DATABASE" -c "SELECT count(*) FROM public.output WHERE label='apr-oracle' AND component IN ('netApr','netApy') AND value < 0; "
```
Expected: `0`.

### Risk / impact

- **Behavioral change**: the `100% performance fee` edge case previously returned `0`; now it returns `gross / 2`. This matches accountant reality (fees cap at 50% of profit) so the old `0` was the bug, but anything downstream that explicitly relied on the old zero will see a non-zero value. `computeNetApr` also now returns `0` instead of a negative for non-positive gross.
- **Rollout order**:
  1. Create a DB snapshot (Neon branch/backup).
  2. Stop the indexer.
  3. Run `bun packages/scripts/src/backfill-apr-oracle-negative-netapr/compute.ts` (writes staged rows to the temp table).
  4. Run `bun packages/scripts/src/backfill-apr-oracle-negative-netapr/upsert.ts` to promote the staged rows and drop the temp table. Note: `upsert.ts` always drops the temp table on success, so any retry requires re-running `compute.ts` first.
  5. Trigger `fanout abis` so the indexer picks up the merged `computeNetApr` floor for future rows.
  6. Restart the indexer.
  7. Trigger the `refresh snapshot` and `refresh list` GitHub Actions workflows so the REST caches pick up the new values.

- **Rollback**: restore the pre-backfill DB snapshot and revert this commit. The old formula would re-introduce negatives for newly ingested rows but the snapshot restore returns historical data to its prior state.

### DIFF
--- a/packages/ingest/abis/yearn/lib/apy.spec.ts
+++ b/packages/ingest/abis/yearn/lib/apy.spec.ts
@@ -210,6 +210,10 @@ describe('abis/yearn/lib/apy', function() {
       expect(computeNetApr(0, { management: 0, performance: 0.20 })).to.equal(0)
     })
 
+    it('returns zero for negative gross APR', function() {
+      expect(computeNetApr(-0.01, { management: 0.0025, performance: 0.10 })).to.equal(0)
+    })
+
     it('floors net APR at half gross APR when fees exceed gross APR', function() {
       const gross = 0.00076638918973244
       expect(computeNetApr(gross, { management: 0.0025, performance: 0.10 })).to.equal(gross / 2)
--- a/packages/ingest/abis/yearn/lib/apy.spec.ts
+++ b/packages/ingest/abis/yearn/lib/apy.spec.ts
@@ -202,12 +202,17 @@ describe('abis/yearn/lib/apy', function() {
       expect(computeNetApr(0.10, { management: 0.02, performance: 0.20 })).to.be.closeTo(0.064, 1e-10)
     })
 
-    it('returns zero when performance fee is 100%', function() {
-      expect(computeNetApr(0.10, { management: 0, performance: 1.0 })).to.equal(0)
+    it('floors net APR at half gross APR when performance fee is 100%', function() {
+      expect(computeNetApr(0.10, { management: 0, performance: 1.0 })).to.equal(0.05)
     })
 
     it('handles zero gross APR', function() {
       expect(computeNetApr(0, { management: 0, performance: 0.20 })).to.equal(0)
+    })
+
+    it('floors net APR at half gross APR when fees exceed gross APR', function() {
+      const gross = 0.00076638918973244
+      expect(computeNetApr(gross, { management: 0.0025, performance: 0.10 })).to.equal(gross / 2)
     })
 
     it('handles no strategies (zero fees)', function() {
--- a/packages/ingest/abis/yearn/lib/apy.ts
+++ b/packages/ingest/abis/yearn/lib/apy.ts
@@ -489,6 +489,10 @@ export async function extractLockedProfit__v3(chainId: number, address: `0x${str
 }
 
 export function computeNetApr(grossApr: number, fees: { management: number; performance: number }): number {
+  // Accountant invariant: total fees cap at 50% of profit, so netApr cannot be
+  // worse than grossApr / 2. For non-positive gross, floor at 0 — a negative
+  // "floor" is meaningless and would let netApr stay negative.
+  if (grossApr <= 0) return 0
   const net = (grossApr - fees.management) * (1 - fees.performance)
   const floor = grossApr / 2
   if (net < floor) return floor
--- a/packages/ingest/abis/yearn/lib/apy.ts
+++ b/packages/ingest/abis/yearn/lib/apy.ts
@@ -489,7 +489,10 @@ export async function extractLockedProfit__v3(chainId: number, address: `0x${str
 }
 
 export function computeNetApr(grossApr: number, fees: { management: number; performance: number }): number {
-  return (grossApr - fees.management) * (1 - fees.performance)
+  const net = (grossApr - fees.management) * (1 - fees.performance)
+  const floor = grossApr / 2
+  if (net < floor) return floor
+  return net
 }
 
 export function computeApy(apr: number): number {
--- a/packages/scripts/src/backfill-apr-oracle-getCurrentApr/compute.ts
+++ b/packages/scripts/src/backfill-apr-oracle-getCurrentApr/compute.ts
@@ -49,6 +49,9 @@ type TempRow = {
   series_time: bigint
 }
 
+// TODO: migrate ensureTempTable + insertTempBatch below onto the shared helpers
+// in ../backfill-shared/tempTable.ts (resetTempTable / insertTempBatch). Left
+// as-is to keep the PR that introduced the shared helpers small.
 async function ensureTempTable() {
   await db.query(`
     CREATE TABLE IF NOT EXISTS ${TEMP_TABLE} (
--- a/packages/scripts/src/backfill-apr-oracle-getCurrentApr/compute.ts
+++ b/packages/scripts/src/backfill-apr-oracle-getCurrentApr/compute.ts
@@ -49,9 +49,6 @@ type TempRow = {
   series_time: bigint
 }
 
-// TODO: migrate ensureTempTable + insertTempBatch below onto the shared helpers
-// in ../backfill-shared/tempTable.ts (resetTempTable / insertTempBatch). Left
-// as-is to keep the PR that introduced the shared helpers small.
 async function ensureTempTable() {
   await db.query(`
     CREATE TABLE IF NOT EXISTS ${TEMP_TABLE} (
--- a/packages/scripts/src/backfill-apr-oracle-getCurrentApr/upsert.ts
+++ b/packages/scripts/src/backfill-apr-oracle-getCurrentApr/upsert.ts
@@ -2,6 +2,7 @@ import 'lib/global'
 
 import db from 'ingest/db'
 import { mq } from 'lib'
+import { parsePromoteArgs, promoteTempTable } from '../backfill-shared/upsert'
 
 /**
  * Phase 2: Promote computed apr-oracle outputs from the temp table into the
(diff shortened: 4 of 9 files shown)


murderteeth (changes requested):
## Summary
Fixes `computeNetApr` to floor at `grossApr / 2` (matching the Yearn accountant invariant that management + performance fees cap at 50% of profit) and adds a compute/upsert pair to backfill the 184 existing historical rows that were stored as negative.

## Issues

- [ ] **packages/scripts/src/backfill-apr-oracle-negative-netapr/{upsert.ts,compute.ts}** — Why does this update introduce so much duplicated code with `backfill-apr-oracle-getCurrentApr/`? `upsert.ts` is ~95% copy-paste (same `parseArgs`, same preview-then-transaction flow, same `42P01` handling, same cleanup — only `TEMP_TABLE` and doc strings differ), and `ensureTempTable` with its `CREATE TABLE` schema is byte-identical across the two `compute.ts` files.

- [ ] **packages/scripts/src/backfill-apr-oracle-negative-netapr/compute.ts:34** — The backfill temp table is not reset between runs, so stale staged rows can be promoted into `output`. `ensureTempTable()` only does `CREATE TABLE IF NOT EXISTS` and keeps any existing contents. If the current source query finds nothing to fix, `compute.ts:100` exits without clearing the table. Meanwhile `upsert.ts:35` reads whatever rows are already present and `upsert.ts:75` promotes all of them. That makes reruns unsafe after negatives were fixed by some other path (or any source-set shrinkage between runs) — prior staged values would silently clobber the current `output` rows. Drop/truncate the temp table at the start of `compute.ts`, or explicitly clear it on the "nothing to backfill" path.

- [ ] **packages/scripts/src/backfill-apr-oracle-negative-netapr/compute.ts:125** — `compute.ts` only backfills rows with `n.value < 0`, but `computeNetApr()` now floors all rows at `grossApr / 2`. Historical rows that are positive but below that floor will remain stale, so post-deploy output can diverge from history. If that's intentional, document it; otherwise broaden the backfill to catch `netApr < grossApr / 2` and `netApy < computeApy(grossApr / 2)`.

- [ ] **packages/scripts/src/backfill-apr-oracle-negative-netapr/README.md:15** — README says "Skips rows … where gross is **non-positive**", but `compute.ts:139` only skips `gross < 0`. Rows with `gross = 0` are staged with `value = 0` (which matches the new formula, so the code is fine). Either change the doc to say "negative" or tighten the check to `<= 0` so they match.

- [ ] **packages/scripts/src/backfill-apr-oracle-negative-netapr/compute.ts** — `db.end()` is called outside any `try/finally`, so a mid-run query failure leaves the pool open. Sibling `backfill-apr-oracle-getCurrentApr/compute.ts` has the same pattern — worth cleaning up in both places.

## Verdict
REQUEST_CHANGES

---

## How This Was Reviewed

matheus1lva:
oh right, the netapr < gross / 2 not necessary imples for <1, which goes to negative, that meant that the net apr being lower than gross /2 - then we keep at 50% only.

matheus1lva (commented):
## Summary

6 commits on `fix/net-apr-floor-half-gross`: floors `computeNetApr` at `grossApr/2` (and at `0` for non-positive gross), plus a two-script backfill (`compute.ts` → `upsert.ts`) that rewrites 184 historical below-floor `netApr`/`netApy` rows across 4 vaults. A shared helper (`backfill-shared/{tempTable,upsert}.ts`) holds the temp-table + promote-drop flow; the sibling `backfill-apr-oracle-getCurrentApr/upsert.ts` was migrated onto it.

Note: GitHub doesn't allow self-approval — posting as a comment-style review. Verdict is **APPROVE**: no blockers.

## Dependencies

No new dependencies. `package.json` unchanged.

## Issues

(No blockers.)

## Suggestions

- **`packages/scripts/src/backfill-shared/upsert.ts:12`** — the "nothing to upsert" early-return prints a message but does not drop the temp table. If `compute.ts` is re-run after an empty upsert, `resetTempTable` will `TRUNCATE` the existing table, so this is benign — but the message `"nothing to upsert. Run compute.ts first."` is slightly misleading in a retry scenario where the temp table is merely empty. Consider `nothing to upsert (temp table empty)`.
- **`packages/scripts/src/backfill-shared/upsert.ts:27`** — `COUNT(DISTINCT (chain_id, address))` returns `bigint` as a string; fine because it's only console-logged, but compare against the `count::int AS n` idiom you adopted on line 7. Tiny consistency nit.
- **`packages/scripts/src/backfill-apr-oracle-negative-netapr/compute.ts:31-37`** — `FindAffectedResult.stageable` duplicates `totalBelowFloor - skippedNoGross - skippedGrossNegative`, and in the happy path always equals `totalStaged`. It's printed as `Staged (gross >= 0):`. Since there's no code path where `stageable` and `totalStaged` legitimately diverge (errors increment `errors`, not `skipped*`), consider dropping `stageable` from the result type and printing `totalStaged` directly — one less piece of bookkeeping to keep in sync.
- **`packages/scripts/src/backfill-apr-oracle-negative-netapr/compute.ts:121-124`** — `STAGED_COMPONENTS` sits between `outputToTempRow` and `replayVault`. Hoisting it next to `TEMP_TABLE` / `CONCURRENCY` at the top of the module would read better.
- **`packages/scripts/src/backfill-shared/tempTable.ts:16-31`** — `CREATE TABLE IF NOT EXISTS` + `TRUNCATE` is fine for the new temp-table name, but if anyone ever renames the shared helper's columns, older in-flight tables will silently `TRUNCATE` with the old schema. Not today's problem; worth a brief code comment saying the temp-table schema is versioned by the table name.
- **`packages/scripts/src/backfill-apr-oracle-negative-netapr/compute.ts:58`** — `component IN ('netApr', 'netApy')` is defensively wide; a below-floor netApy always implies a below-floor netApr at the same `series_time` (since netApy is monotonic in netApr), so restricting to `'netApr'` would return the same set of affected series_times after dedup. Not wrong — just worth a one-line comment that the inclusion is belt-and-braces.
- **PR description — Rollout step 4** correctly notes the temp table is dropped. The dry-run preview step was removed, so any partial-trust run now depends on the `sample rows:` / `distinct vaults:` lines inside `upsert.ts`. That's acceptable but worth knowing: the only preview is embedded inline, not a separate command.

## Verification

- Lint: `bun --filter ingest lint` and `bun --filter lib lint` → 0 errors (all warnings pre-existing).
- Unit tests: the `computeNetApr` cases are nested inside `describe('abis/yearn/lib/apy')` whose `beforeAll` hits Postgres, which isn't reachable from this shell. By manual trace all six cases (including the new `returns zero for negative gross APR`) pass against `apy.ts:491-500`. Please run `bun --filter ingest test` with `make dev` up before merging.
- No UI changes; Playwright not applicable.

## Verdict

**APPROVE** (posted as COMMENT — self-approval blocked by GitHub). No blockers; suggestions are all polish / readability.

---

## How This Was Reviewed
This review was conducted using the [review-local skill](https://github.com/yearn/webops/blob/main/skills/review-local/SKILL.md).

matheus1lva:
## Summary

Fixes `computeNetApr` to enforce the accountant's "fees cap at 50% of profit" invariant — floors net APR at `grossApr / 2` for positive gross and at `0` for non-positive gross. Adds a backfill (`compute.ts` + `upsert.ts`) that stages rewrites of 184 historical below-floor `netApr`/`netApy` rows across 4 vaults to a temp table, then promotes them. Extracts a shared `promoteTempTable` helper and migrates the sibling `getCurrentApr` backfill onto it.

## Issues

None blocking. A few observations:

- **`packages/scripts/src/backfill-shared/tempTable.ts:16` / `:33`** — `resetTempTable(name)` and `insertTempBatch(name, ...)` interpolate `name` directly into SQL. Safe today (hardcoded call sites) but worth a one-line note that `name` is expected to be a hardcoded identifier, not user input, if this helper picks up more callers.
- **`packages/scripts/src/backfill-apr-oracle-negative-netapr/compute.ts:104`** — `outputToTempRow` sets both `block_time` and `series_time` to the series-bucket timestamp, overwriting the original block-aligned `block_time` on rewritten rows. This is called out in the PR description, but worth confirming no REST/GraphQL consumer reads `block_time` for `apr-oracle` rows and would misinterpret the shifted values.
- **Partial shared-helper migration**: `backfill-apr-oracle-getCurrentApr/compute.ts` still has its own local `ensureTempTable` and `insertTempBatch`, duplicating logic now in `backfill-shared/tempTable.ts`. Not in scope to migrate here, but leaving a TODO or follow-up would prevent drift.

## Suggestions

- The `100% performance fee` test name change is correct, but the behavior shift from `0` → `gross/2` for that specific case is semantic. If any downstream consumer treats `0` as a sentinel for "fee captured all profit," it'll now see a non-zero value. PR description already flags this.
- `findAffected` anchoring on `netApr` alone and letting the replay recompute `netApy` is a good call — comment explains the non-linearity of `computeApy` well.

## Verdict

APPROVE (with the `block_time` observation worth a sanity check before running the backfill in prod).

---

## How This Was Reviewed

murderteeth (commented):
lgtm except `bun packages/scripts/src/backfill-apr-oracle-negative-netapr/compute.ts` seems to hang right here,

Image

does that matter?

also regarding "..i'd prefer to have them "duplicate" where i pull the dataset already in the compute rather than generate in a json" i think that's totally fine!

matheus1lva:
> lgtm except `bun packages/scripts/src/backfill-apr-oracle-negative-netapr/compute.ts` seems to hang right here,
> 
> Image
> does that matter?
> 
> also regarding "..i'd prefer to have them "duplicate" where i pull the dataset already in the compute rather than generate in a json" i think that's totally fine!

yeah the first time i ran this, it took some significant time, so it's going. Im rerunning against another fork now given this was udated. Let's see!

matheus1lva:
Ok i've reran it, went succesfully, yeah it was a bit slower. I added more concurrency! @murderteeth , just wiating to get merged to run the upsert!

murderteeth (approved):

PR #399, opened 2026-05-06
feat: add price service as fallback for price fetching

matheus1lva:
## Summary

- Adds \`fetchPriceServiceUsd\` as a last-resort fallback when Lens, yPrice, and yDaemon all return nothing
- Calls \`prices.yearn.dev/api/prices/batchHistorical\` with \`PRICE_SERVICE_API_KEY\` — skipped silently if key is absent
- Covers chains not previously reachable (sonic, berachain, katana) in addition to existing ones

### Setup

- [x] Copy \`.env.example\` → \`.env\`, set \`PRICE_SERVICE_API_KEY\` to a valid key
- [x] \`make dev\`

### Unit test — fallback fires for BOLD (Liquity v2, unsupported by Lens)

```bash
PRICE_SERVICE_API_KEY=<key> bun --filter ingest test --grep "priceservice"
```
Expected: test passes, \`priceSource === 'priceservice'\`, \`priceUsd > 0\`

### Integration via fanout — trigger a price fetch for a BOLD vault

Set up \`config/abis.local.yaml\` pointing at a vault whose underlying asset is BOLD (\`0x0655977FEb2f289A4aB78af67BAB0d17aAb84367\`):

```yaml
# config/abis.local.yaml
cron:
  name: AbiFanout
  queue: fanout
  job: abis
  schedule: '*/15 * * * *'
  start: false

abis:
  - abiPath: 'yearn/3/vault'
    sources: [
      { chainId: 1, address: '0x9F4330700a36B29952869fac9b33f45EEdd8A3d8', inceptBlock: 22569667 }
    ]
```

```yaml
# config/manuals.local.yaml
manuals:
  - chainId: 1
    address: '0x9F4330700a36B29952869fac9b33f45EEdd8A3d8'
    label: 'vault'
    defaults: {
      erc4626: true,
      v3: true,
      yearn: true,
      asset: '0x6440f144b7e50D6a8439336510312d2F54beB01D',
      inceptBlock: 22569667,
    }
```

- [x] In terminal UI → `ingest` → `fanout abis`
- [x] Query the price table:
```sql
SELECT price_source, price_usd FROM price
WHERE address = '0x6440f144b7e50D6a8439336510312d2F54beB01D'
ORDER BY block_number DESC LIMIT 5;
```
Expected: `price_source = 'priceservice'`, `price_usd > 0`

### Graceful skip when key is absent

- [ ] Unset `PRICE_SERVICE_API_KEY` in `.env`, restart ingest
- [ ] Re-run the unit test:
```bash
bun --filter ingest test prices.spec.ts
```
Expected: test is skipped (not failed), no crash in ingest logs


murderteeth (changes requested):
- use envar with default for `https://prices.yearn.dev`
- unit tests return:
`error TS5023: Unknown compiler option '--grep'`
- query is wrong, lcase address wont match anything:
`SELECT price_source, price_usd FROM price
WHERE address = '0x0655977feb2f289a4ab78af67bab0d17aab84367'
ORDER BY block_number DESC LIMIT 5;`
- i see price gaps, it looks like for particular days
Image
Image

matheus1lva:
That was wierd, it returned correct when running the tests for mocha, let me also check if the price service returns stuff when requested at mid day.

matheus1lva:
@murderteeth so i did an update on mocha - which was painfully unstable to run the tests - better now `bun --filter ingest test prices.spec.ts` should do the job. And for some gaps on day past today - yeah im debugging that in the price service.
And for today as well - because we only record the EOD - i guess we would still see 0 in certain parts of the day.

matheus1lva:
All review points addressed in latest commit:

- **PRICE_SERVICE_URL default**: now defaults to `https://prices.yearn.dev` when env unset.
- fixed the test setup script which now honors any new path you pass to it, ie: `bun run --filter ingest test prices.spec.ts`

murderteeth (changes requested):
## Summary
Adds `fetchPriceServiceUsd` as a last-resort fallback after Lens / yPrice.

## Issues

- [ ] **packages/ingest/prices.ts:147** — The `getAddress(token)` change is on the read path only; every write path (`fetchLensPriceUsd:169`, `fetchYPriceUsd:125`, `fetchYDaemonPriceUsd:216`, the new `fetchPriceServiceUsd:100`, and the `empty` fallback at `:64`) still inserts `address: token` in caller-provided casing, and `PriceSchema.address` is `zhexstring` (no transform) so the casing survives all the way to the case-sensitive `price.address` Postgres column. A lowercase caller therefore writes a lowercase row, then the new checksum lookup misses it and the fetchers run again, allowing duplicate PK rows in `(chain_id, address, block_number)` with different casing. Recommend normalizing once near the top of `fetchErc20PriceUsd` / `__fetchErc20PriceUsd` (single `token = getAddress(token)` and let everything downstream inherit it), or normalize every `PriceSchema.parse({ address })` write site in addition to the read path. Just dropping the new `getAddress(token)` would put back the original lowercase-query problem flagged earlier, so that's not the right rollback.

- [ ] I still get errors running tests (am i doing it right?)
image


## Verdict
REQUEST_CHANGES

---

## How This Was Reviewed

matheus1lva:
> ```shell
> ```shell
> bun --filter ingest test prices.spec.ts
> ```
> 
> 
>     
>       
>     
> 
>       
>     
> 
>     
>   
> ```

I updated a UX issue on the way that mocha handles these so if you run with:
```
bun --filter ingest test prices.spec.ts
```

that should work

murderteeth (approved):
## Summary
Adds `fetchPriceServiceUsd` as a last-resort fallback after Lens / yPrice / yDaemon and addresses prior-review feedback (env var default, test-runner args, address casing on writes).

## Issues
- None blocking. Prior casing concern at `fetchErc20PriceUsd` is resolved by normalizing once at entry (`prices.ts:19`); checksummed `token` propagates through `__fetchErc20PriceUsd` into every sub-fetcher's `PriceSchema.parse({ address: token })` write site, the cache key, and the `empty` fallback.

## Verdict
APPROVE

---

## How This Was Reviewed

PR #406, opened 2026-05-13
perf(db): cut CPU ~78% via evmlog/output indexes + hot query rewrites

matheus1lva:
## Summary

Neon CPU roughly doubled since the start of the year, and Neon changed CPU billing. The 2026-05-13 `pg_stat_statements` snapshot found three queries driving most of the cost, all reads against `evmlog` (5.27 M rows / 6.9 GB) and the `output` hypertable (27.5 M rows / 128 Timescale chunks), all from missing indexes or a missing `series_time` predicate.

Full breakdown and ranked plan: [`docs/cpu-cost-analysis.md`](./docs/cpu-cost-analysis.md).

This is the last of the three to land. The other two already shipped:

| baseline query | baseline | live 2026-06-10 | status |
|---|---|---|---|
| #1 `projectDebtAllocator` (evmlog full scan) | 9.4 s, 39.4% | out of top 10, evmlog `seq_scan` 0 | already fixed in prod by `idx_evmlog_chain_sig_block` (chain+signature index shipped separately) |
| #2 `fetchStrategyPerformance` | 34 s, 35.0% | **40.7 s, 68.6%** | **this PR** |
| #3 timeseries fanout (`DISTINCT block_time`) | 191 ms, 8.7% | 7.9 ms, 0.59% | fixed by #413 (96% / 24x), old query retired |

The remaining lever is the #2 query, now **68.6% of all DB exec time**. It reads 64 TB of shared blocks over 43 k calls (~1.5 GB/call) because it scans all 128 chunks per call.

### What this PR changes

1. **`fetchStrategyPerformance` rewrite** (`vault/snapshot/hook.ts`): bound the CTE by `series_time >= now() - interval '30 days'` for chunk exclusion, use `MAX(series_time)` and join on `series_time`, and repeat the chain_id/address/label/series_time predicates on the outer `output o` scan so exclusion holds under any join plan (nested loop or hash join). The query stays CTE+JOIN on `MAX(series_time)`; a `DISTINCT ON` variant broke per-component coherence, reverted in 8c470ee.
2. **`output(chain_id, address, label, series_time DESC)` index** backs that rewrite. The existing `output(…, block_time DESC)` index is keyed on `block_time`, which does not give chunk exclusion; `series_time` does.
3. **`projectDebtAllocator` predicate** (`hook.ts`): added `args ? 'vault'` so the query matches the partial index below.
4. **Partial expr index** on `evmlog (chain_id, signature, (args->>'vault')) WHERE args ? 'vault'` for the `projectDebtAllocator` vault seek.
5. **`fanout/timeseries.ts`** comment only; the bounded `series_time` fanout rewrite shipped in #413.

### Schema changes (migration `20260513233437-cpu-indexes`)

```sql
CREATE INDEX IF NOT EXISTS evmlog_idx_chain_signature_args_vault
  ON evmlog (chain_id, signature, (args->>'vault'))
  WHERE args ? 'vault';

CREATE INDEX IF NOT EXISTS idx_output_chain_address_label_series_time
  ON output (chain_id, address, label, series_time DESC);
```

The generic `evmlog (chain_id, signature, block_number DESC, …)` index this migration first proposed is dropped: prod already has it as `idx_evmlog_chain_sig_block`, and `IF NOT EXISTS` would not catch the different name, so the migration would build a redundant 6.9 GB duplicate. Deployment: build both `CONCURRENTLY` against prod first (the partial evmlog index is small; the output index is on a hypertable), then run the migration as a no-op.

### Live measurement (2026-06-10, prod `green-frog-37871507`)

`pg_stat_statements` window 2026-06-09 04:02Z to 2026-06-10 16:25Z (~1.5 days). Top DB-time consumers right now:

| % exec | CPU-hours | calls | mean | query |
|---|---|---|---|---|
| 68.6% | 486.1 | 43,022 | 40.7 s | `fetchStrategyPerformance` (old `MAX(block_time)`, no series_time bound) |
| 11.8% | 83.4 | 1.03 M | 292 ms | estimated-apr resolver |
| 5.9% | 41.8 | 431 k | 349 ms | apy `MAX(CASE…)` resolver |
| 3.8% | 26.6 | 342 k | 280 ms | apr `MAX(CASE…)` resolver |
| 2.8% | 19.6 | 177 k | 397 ms | latest-snapshot resolver |

`fetchStrategyPerformance` is the #1 cost by 6x. The rewrite plus the `series_time` index removes the all-chunk scan.

### #413 measured gain (review response)

#413 ("fanout timestamp bounded discovery") merged on 2026-05-29 (`dbe9622`); this branch picks it up through the main merge (`dcc0ed3`), and the only `timeseries.ts` delta here is a 4-line comment. #413 rewrote the timeseries fanout discovery from `SELECT DISTINCT block_time` over all history to a `series_time BETWEEN` bounded scan.

**One number:** the fanout discovery query's mean cost fell from **191 ms/call to 7.9 ms/call, a 96% reduction (24x faster)**. Source: 2026-05-13 baseline (191 ms mean) vs live 2026-06-10 (7.91 ms mean over 1.92 M calls). The old unbounded query has 0 calls remaining, and the bounded form is now 0.59% of exec time, down from 8.7%.

### CPU chart note

The aggregate compute-CPU chart does not isolate #413: its ~8% slice is buried under backfill bursts (compute pins the 4 CU ceiling, suspend_timeout 0) and the still-unfixed 68.6% query. The clean #413 signal is the per-call 191 to 7.9 ms. The visible step-down on the chart will come when this PR deploys and removes the 68.6% query.

### Out of scope (follow-ups)

Listed in `docs/cpu-cost-analysis.md` under the fix plan, in materiality order:

- Cache `projectDebtAllocator` result in the vault snapshot row.
- Add `series_time` lower bound to the apr/apy `MAX(CASE WHEN …)` resolvers (the 11.8% / 5.9% / 3.8% consumers above).
- Resolver-level cache for `timeseries` / `tvls` GQL queries.
- Kill the `event_name COUNT(*)` probe in `packages/ingest/probe/index.ts` (70 s per call).
- Tune Neon compute on idle/read replica branches.

### Validation status

- `hook.spec.ts` unit coverage and migration up/down idempotency: done locally.
- Remaining: `EXPLAIN (ANALYZE, BUFFERS)` of the rewritten `fetchStrategyPerformance` on a prod fork to confirm chunk exclusion under a forced hash join, then a post-deploy `pg_stat_statements` re-read for the real CPU-seconds drop.

> On the one-commit-per-change feedback: this branch is already pushed and shared, so I'm not rewriting its history. New commits from here are scoped per change.

### Setup

- [x] Confirm `.env` `POSTGRES_*` points at a dev branch (not prod).
- [ ] `make dev` starts redis, postgres, ingest, web.

### Migration

- [ ] Apply migration on the dev DB:
  ```
  bun --filter db migrate up
  ```
  Expected: two `CREATE INDEX` statements run cleanly; subsequent runs are no-ops.
- [x] Verify indexes exist:
  ```sql
  SELECT indexname FROM pg_indexes
  WHERE tablename IN ('evmlog','output')
    AND indexname IN (
      'evmlog_idx_chain_signature_args_vault',
      'idx_output_chain_address_label_series_time'
    );
  ```
  Expected: 2 rows.
- [x] Rollback path works:
  ```
  bun --filter db migrate down -c 1
  ```
  Expected: 2 indexes dropped. Then `migrate up` to re-apply.

### Hot query #2: `fetchStrategyPerformance` (the target)

- [x] After a fanout run, spot-check a vault's snapshot via GraphQL and confirm strategy performance fields (`oracle.apr`, `historical.weeklyNet`, etc.) populate as before.
- [x] `EXPLAIN ANALYZE` the rewritten query against a prod fork and confirm BOTH the CTE and the outer scan use `idx_output_chain_address_label_series_time` with chunk exclusion (Timescale `Chunks excluded`), under a forced hash join too (`SET enable_nestloop = off`) to prove the outer predicates hold.

### Hot query #1: `projectDebtAllocator` (already fast in prod; verify no regression)

- [ ] Pick a v3 vault and confirm the lookup uses an index and returns in < 20 ms:
  ```sql
  EXPLAIN (ANALYZE, BUFFERS) SELECT args FROM evmlog
  WHERE chain_id = 1
    AND signature = '<NewDebtAllocator topic>'
    AND args ? 'vault' AND args->>'vault' = '<vault addr>'
  ORDER BY block_number DESC, log_index DESC LIMIT 1;
  ```
  Expected: Index Scan on `evmlog_idx_chain_signature_args_vault` (or the existing `idx_evmlog_chain_sig_block`), no seq scan.
- [ ] Terminal UI → ingest → `fanout abis` and confirm v3 vault snapshots complete without errors.

### Hot query #3: Timeseries fanout (shipped in #413)

- [ ] Terminal UI → ingest → `fanout abis` (timeseries). Confirm jobs enqueue identically to before for a known vault: same set of `endOfDay` timestamps queued.
- [ ] A vault with a partially-backfilled history should enqueue the *missing* days only. Compare the set of `mq.add` calls to a baseline if reproducible.

### Regression: replays

- [ ] `fanout replays` from terminal UI; confirm replays still complete and produce the same outputs.

### Cleanup

- [x] `make down`


matheus1lva:
**Production measurement — #413 landed, this PR is the logical next cut.**

Pulled live `pg_stat_statements` from Neon (`kong`, 40.6d window since the Apr 28 reset). #413's fanout fix (Target #1) is confirmed:

| fanout query | before (`DISTINCT block_time`) | after (`series_time` bounded) |
|---|---|---|
| rows / call | 513.8 | **14.9** (−97.1%) |
| mean exec | 199.9 ms | **7.25 ms** (−96.4%, 27.6×) |
| top-offender rank | top row-offender | **dropped out of top 12** (~0.09% CPU) |

≈82 GB/mo egress + ≈57 CPU-days/mo saved on that one query. The methodology holds, so the next two offenders are now the whole game.

**What this PR fixes — the current live top 2 (~72% of DB CPU):**

| live rank | % CPU | mean | query | this PR |
|---|---|---|---|---|
| #1 | 36.6% | 36.5 s | `fetchStrategyPerformance` CTE (`MAX(block_time)`, no time bound → full hypertable scan) | rewrite → `DISTINCT ON` + `series_time >= now()-30d`, chunk-prunable |
| #2 | 35.8% | 8.65 s | `args FROM evmlog … args->>'vault'` (PK can't serve `(chain_id, signature)` → full scan) | new `evmlog(chain_id, signature, …)` + partial `args->>'vault'` index |

Note: this PR's item #3 (fanout `DISTINCT block_time`) is already done — shipped in #413 — so the realizable win here is items #1+#2. Expect mean exec on the strategy-perf query to collapse from 36.5s once chunk pruning kicks in (same mechanism that took fanout 199.9ms→7.25ms).

murderteeth (commented):
- explain what the net measured cpu usage gain was from merging 413 in one clear number\percentage so that it's easy to justify more optimizations
- why are 413's changes included in this branch? that was not clear to me
- has this migration been tested using a fork? include notes in PR
- this would be easier\faster to review if each change were in its own commit with its own commit notes

notes from my bot:
- [ ] packages/ingest/abis/yearn/3/vault/snapshot/hook.ts:418-424 — The outer scan of output o has only o.chain_id = $1 as a direct predicate; address, label, and series_time arrive only through the join
  with latest_times. Chunk exclusion on the outer side then depends on the planner choosing a nested loop with runtime exclusion — if it picks a hash join, the outer side degrades back to scanning every chunk
  for the chain, which is the exact pathology this PR fixes (query #2, 35% of CPU). The earlier DISTINCT ON version had the direct predicates; the coherence rewrite (8c470ee) dropped them. Add them back to the
  outer query — redundant with the join, but they guarantee static chunk exclusion and use of idx_output_chain_address_label_series_time:
  WHERE o.chain_id = $1
    AND o.address = ANY($2)
    AND o.label = ANY($3)
    AND o.series_time >= now() - interval '30 days'
- [ ] PR description — Stale relative to head: it says fetchStrategyPerformance was rewritten to DISTINCT ON (it's back to CTE+JOIN on MAX(series_time) since 8c470ee, intentionally, for component coherence),
  and lists the timeseries.ts fanout rewrite as a code change (that shipped in #413; this PR now only adds a comment there). Since the test plan and deployment notes are written against the description,
  update it before merge.

matheus1lva:
1st: the cart itself after the merge of the first pr

CleanShot 2026-06-10 at 13 35 38@2x

Which ingested the results of the chart from neon onto ai and:

## Live measurement (2026-06-10, prod `green-frog-37871507`)

`pg_stat_statements` window 2026-06-09 04:02Z → 2026-06-10 16:25Z (~1.5 days).
Tables: evmlog 5.27 M rows / 6.9 GB; output 27.5 M rows across 128 Timescale chunks.

Top DB-time consumers right now:

| % exec | CPU-hours | calls | mean | query |
|---|---|---|---|---|
| 68.6% | 486.1 | 43,022 | 40.7 s | `fetchStrategyPerformance` (old `MAX(block_time)`, no series_time bound) |
| 11.8% | 83.4 | 1.03 M | 292 ms | estimated-apr resolver |
| 5.9% | 41.8 | 431 k | 349 ms | apy `MAX(CASE…)` resolver |
| 3.8% | 26.6 | 342 k | 280 ms | apr `MAX(CASE…)` resolver |
| 2.8% | 19.6 | 177 k | 397 ms | latest-snapshot resolver |

`fetchStrategyPerformance` reads 64 TB of shared blocks over those 43 k calls
(~1.5 GB/call) because it scans all 128 chunks per call. It is the current #1
cost by 6x and is exactly what this PR rewrites.

Per-query progress (baseline 2026-05-13 vs live 2026-06-10):

| query | baseline | live | status |
|---|---|---|---|
| timeseries fanout (#3) | 191 ms, 8.7% | 7.9 ms, 0.59% | fixed by #413 (96% / 24x), old query retired |
| projectDebtAllocator (#1) | 9.4 s, 39.4% | out of top 10 | fixed by `idx_evmlog_chain_sig_block`, evmlog seq_scan 0 |
| fetchStrategyPerformance (#2) | 34 s, 35.0% | 40.7 s, 68.6% | this PR |

With #1 and #3 already landed, this PR's lever is the 68.6% query.

## CPU chart note (2026-05-27 → 06-03)
The aggregate compute-CPU chart does not isolate #413: its ~8% slice is buried
under backfill bursts (compute pins the 4-CU ceiling, suspend_timeout 0) and the
still-unfixed 68.6% query. The clean #413 signal is the per-call 191 → 7.9 ms.

Plus i also added an index while looking at the most slow queries through this time which also helped decrease such.

murderteeth (approved):
lgtm, but please address these notes before rollout,

> The aggregate compute-CPU chart does not isolate #413: its ~8% slice is buried under backfill bursts (compute pins the 4-CU ceiling

so we don't have a clear measure of 413's net impact on cpu usage. that's ok, but how will we measure this PR's impact. does neon's api have anything we can make our own charts from? the chart you screen captured shows CPU utilization, not CPU usage hours. if neon api has cpu usage, lets try making a chart or table from that instead

> Plus i also added an index while looking at the most slow queries through this time which also helped decrease such.

did you add this index through a migration or did you create it directly on the database? if you make it directly, please go ahead and add it to this PRs migration.

> from the test plan, `bun --filter db migrate up`

not needed, `make dev` already does this. but yes do this during rollout even if you manually create the indexes. otherwise prod's schema will get out of sync.

---

i also timed the index migrations. evmlog_idx_chain_signature_args_vault takes about 30s. idx_output_chain_address_label_series_time takes 2 minutes 30 seconds. please keep in mind during rollout.

matheus1lva:
> lgtm, but please address these notes before rollout,
> 
> > The aggregate compute-CPU chart does not isolate #413: its ~8% slice is buried under backfill bursts (compute pins the 4-CU ceiling
> 
> so we don't have a clear measure of 413's net impact on cpu usage. that's ok, but how will we measure this PR's impact. does neon's api have anything we can make our own charts from? the chart you screen captured shows CPU utilization, not CPU usage hours. if neon api has cpu usage, lets try making a chart or table from that instead
> 
> > Plus i also added an index while looking at the most slow queries through this time which also helped decrease such.
> 
> did you add this index through a migration or did you create it directly on the database? if you make it directly, please go ahead and add it to this PRs migration.
> 
> > from the test plan, `bun --filter db migrate up`
> 
> not needed, `make dev` already does this. but yes do this during rollout even if you manually create the indexes. otherwise prod's schema will get out of sync.
> 
> i also timed the index migrations. evmlog_idx_chain_signature_args_vault takes about 30s. idx_output_chain_address_label_series_time takes 2 minutes 30 seconds. please keep in mind during rollout.

yep no it doesnt have CPU utilization chart in their dashboard, im gonna try to find if theres any hidden api or something before mergin this

PR #400, opened 2026-05-07
first phase of db improvements

matheus1lva:
### Summary
Implements the first phase of the Neon egress reduction plan from the HackMD. The branch reduces rows/bytes returned by hot ingest queries without adding a cache layer.

Main changes:
- Bound fanout timestamp discovery to the requested series window using `series_time`.
- Replace the strategy-performance `latest_times` CTE/JOIN with a bounded `DISTINCT ON (address, label, component)` scan.
- Fetch only `thing.defaults` in hot timeseries hooks instead of full `thing` rows.
- Push `things.get()` equality / inequality filters into SQL and keep semver filtering in JS.
- Bound current APY/APR MAX pivots to the last 7 days.
- Narrow targeted `evmlog.args` reads to the fields each caller actually uses.

### Expected savings
These are directional estimates from the HackMD diagnosis, not proof of production impact. The real pass/fail check is the post-deploy `pg_stat_statements` resample after about 3 days.

- Fanout timestamp discovery was the biggest measured row offender: about 4M calls / 3.5d, 2B rows, and 16 GB estimated egress. Bounding it to the requested window should plausibly remove 99%+ of returned rows for that query shape.
- Strategy performance fetch had the largest CPU cost: about 91K calls / 3.5d and roughly 35 days of database CPU over that window. The bounded `DISTINCT ON` scan should turn the full-history CTE/JOIN into a recent-window lookup.
- `thing` and `evmlog` projection changes reduce JSONB bytes returned at hot call sites. These are expected to help most where callers only need a few fields from large `defaults` / `args` blobs.
- Overall expectation: targeted queryids should drop by at least 80% in returned rows, with the fanout query expected at 99%+. Invoice-level savings still need the production measurement window because JSONB/TOAST wire bytes are not fully captured by local tests.
### How to review
Start with:
- `packages/ingest/fanout/timeseries.ts`
- `packages/ingest/abis/yearn/3/vault/snapshot/hook.ts`
- `packages/ingest/things.ts`
- `packages/ingest/helpers/apy-apr.ts`

Then skim the small hook updates under `packages/ingest/abis/yearn/**` and `packages/ingest/abis/erc4626/**` to confirm shape-preserving query projection changes.

Behavior should stay the same for normal/current data. The intentional tradeoff is that current strategy performance and APY/APR helpers now ignore stale output outside their bounded lookback windows instead of returning old historical values.

### Test plan
Automated checks run:
```bash
bun run --filter ingest lint
[...]

PR #413, opened 2026-05-27
fix: test runner cleanup + fanout timestamp bounded discovery

matheus1lva:
## Summary

Split from #400 per review feedback — smaller, independently measurable changes.

### Test runner cleanup
- Fix mocha arg parsing (split `--timeout` from value)
- Default to `**/*.spec.ts` when no spec args provided
- Remove `describe.only` from `strider.spec.ts` (was still there on the old branch)
- Remove `it.only` from strider rollback test
- Fix `hook.spec.ts` blockTime to be recent (within 7-day lookback window)

### Fanout timestamp discovery
- Replace unbounded `DISTINCT block_time` scan with `series_time` bounded query on the output table, scoped to the fanout window
- Maps to the top `pg_stat_statements` offender — the 99%+ row-reduction win
- Deploy and measure before adding more changes

## From
Split from #400 (PR 1+2 per review request)

## Follow-ups (surfaced during review)
- #415 — broken `types.spec.ts` fixture (pre-existing, unmasked by the `.only` removal)
- #416 — `bun --filter ingest test` silently skips most specs (shell globstar default-off; pre-existing in `packages/ingest/run-tests.ts`)

### DIFF
--- a/packages/ingest/abis/yearn/2/strategy/snapshot/hook.spec.ts
+++ b/packages/ingest/abis/yearn/2/strategy/snapshot/hook.spec.ts
@@ -27,7 +27,8 @@ describe('abis/yearn/2/strategy/snapshot/hook', function() {
   it('extracts estimated apr', async function() {
     const chainId = 1337
     const address = '0x1000000000000000000000000000000000000000'
-    const blockTime = 1000n
+    // must be recent: getLatestEstimatedApr filters block_time > NOW() - 7 days
+    const blockTime = BigInt(Math.floor(Date.now() / 1000))
     const blockNumber = 1000n
 
     const outputData = {
--- a/packages/ingest/fanout/timeseries.ts
+++ b/packages/ingest/fanout/timeseries.ts
@@ -28,12 +28,13 @@ export default class TimeseriesFanout {
       const end = endOfDay(await getBlockTime(chainId, to))
 
       const computed = (await db.query(`
-      SELECT DISTINCT block_time
+      SELECT DISTINCT FLOOR(EXTRACT(EPOCH FROM series_time))::bigint AS series_time
       FROM output
       WHERE chain_id = $1 AND address = $2 AND label = $3
-      ORDER BY block_time ASC`,
-      [chainId, address, outputLabel]))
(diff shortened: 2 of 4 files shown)


murderteeth (approved):
## Summary
Two unrelated fixes bundled, split from #400:
- Test runner: fix mocha `--timeout` arg splitting, default to `**/*.spec.ts` fallback, drop `describe.only`/`it.only` from strider, bump `hook.spec.ts` blockTime into the 7-day lookback window.
- Fanout timestamp discovery: replace unbounded `DISTINCT block_time` scan with a `series_time` range-bounded scan on `output`, returning epoch seconds directly. Aligns with the hypertable partition column for chunk pruning.

## Issues
None blocking.

## Follow-ups (not blockers, separate from this PR)
- **`packages/lib/types.spec.ts` is broken on disk** — Removing the `.only` calls in `strider.spec.ts` correctly unmasks every other lib spec, which surfaces a pre-existing failure: `OutputSchema.parse({ address: '0x', ... })` is rejected by `EvmAddressSchema → viem.getAddress`. The test fixture has `address: '0x'`, which has never been a valid checksum address. Either fix the fixture (use a real address) or loosen the schema. Worth fixing soon so `bun --filter lib test` goes back to exit 0.
- **`bun --filter ingest test` discovery is broken (pre-existing)** — `packages/ingest/run-tests.ts` (unchanged here) uses the same `['**/*.spec.ts']` fallback with `spawn(..., { shell: true })`. Bash's `globstar` is off by default, so `**` collapses to `*` and only depth-2 specs are discovered (`abis/index.spec.ts`, `extract/evmlogs.spec.ts`). Top-level (`things.spec.ts`, `prices.spec.ts`, `abiutil.spec.ts`) and depth-3+ specs are silently skipped — including the `hook.spec.ts` blockTime fix in this PR at `abis/yearn/2/strategy/snapshot/hook.spec.ts`. The fix itself is correct, but the runner doesn't exercise it. Options: drop `shell: true` and let mocha's own glob handle `**`, prefix the shell with `shopt -s globstar`, or pre-expand with a JS glob lib.

## Verdict
APPROVE

---

## How This Was Reviewed

PR #417, opened 2026-05-30
fix: cut indexer RPC egress via block cache + interpolation height search

matheus1lva:
## Problem

The indexer made ~1000 `eth_getBlockByNumber` POSTs/min on Render. A throwaway byte-meter on a prod fork attributed **the overwhelming majority of RPC _response_ bytes to `eth_getBlockByNumber`**.

**Billing correction (per review).** [Render bills only _outbound_ bandwidth](https://render.com/docs/outbound-bandwidth); inbound is explicitly not billed. `eth_getBlockByNumber` _response_ bytes are **inbound**, so this PR does **not** cut the billed egress. It cuts RPC call volume (provider-side load plus our inbound traffic). The billed ~500GB/mo is **outbound** and remains unexplained by getBlock. Outbound measured small (~tens of GB/mo) in these short runs, so it needs its own investigation. Next target for the billed direction: **multicall / snapshot `eth_call` request calldata**, which is outbound.

Two compounding causes live in `packages/lib/blocks.ts`:
- `getBlock` fetches the **full block (~16KB)** but keeps only `{number, timestamp}`, cached **10s**, even for immutable historical blocks.
- `estimateHeightManual` resolves `blockTime → blockNumber` by **binary search over `[0, head]` (~24 `getBlock`)**. Every timeseries hook does this (apy 3×), and the deep mids are distinct per target, so the cache misses nearly every time.

## Fix

1. **TTLs derived from the indexing workload** (not from block immutability): head (no `blockNumber`) cached **15m** (one indexing cycle), was 10s; historical blocks **1 day**, was 30d. Head pinning gives every job in a cycle one consistent per-chain height and collapses the `estimateHeight` storm to deterministic cache hits; historical only needs to outlive the backfill reuse window (sweeps re-probe the same day-boundary blocks for hours, then never again). The cache holds only `{number,timestamp}`, so a shallow reorg shifts a recent timestamp by ~seconds, with no `endOfDay` bucket impact.
2. **Interpolation search** instead of binary: probe by linear time→block estimate, ~O(log log n) → **~4 `getBlock` vs ~24**, same bracketed result. Div-by-zero falls back to bisect; clamp guarantees termination.
3. **Block 0 routing**: `getBlock(chainId, 0n)` no longer falls through the falsy-bigint check (returning head); `useArchiveNode` likewise treats `0n` as the deepest historical block and routes it to an archive node.

## Measurement: caveats first

These numbers count **inbound `eth_getBlockByNumber` response bytes / call volume**, i.e. RPC load, **not** the billed outbound egress (see billing correction above). They come from a **throwaway byte-meter** (wraps `fetch`, tallies bytes per RPC method) over **short, uncontrolled runs on a forked, backfill-heavy prod DB**. This is not a controlled benchmark: run lengths differ, caches start cold, block-size mix varies. The GB/mo column is the meter's own `bytes/min × month` extrapolation, order-of-magnitude, not precise.

Raw `eth_getBlockByNumber` capture per run:

| stage | run len | getBlock POSTs | POSTs/min | meter GB/mo (proj.) |
|---|---|---|---|---|
| baseline | 27.6 min | 28411 | ~1029 | ~711 |
| cache only | 10.9 min | 9647 | ~885 | ~444 |
| cache + interpolation | 15.8 min | 6527 | ~413 | ~167 |

**Most trustworthy signal** (within-run, run-length-independent): `getBlock` calls per `eth_call` (hook fire) dropped **~2.2 → ~0.8**, and `getBlock` POSTs/min **~1029 → ~413** (~60% fewer RPCs). The GB/mo projections (~700 → ~170) point the same direction but stay noisy. POSTs/min and GB/mo don't agree on the exact %, which is why I'm not claiming a precise figure.

Residual is a backfill floor (each distinct day needs one uncacheable answer block); a caught-up steady-state prod should sit lower.

## Test plan

- [x] `tsc --noEmit` clean for `blocks.ts`
- [x] Inbound getBlock call reduction observed before/after on a forked prod DB (rough, above)
- [x] `blocks.spec.ts` asserts block 0 resolves as a historical block **and** routes to an archive node
- [ ] Reviewer: confirm interpolation returns sane heights vs prior binary search for a few known (chain, timestamp) pairs
- [ ] Separate follow-up: meter **outbound** multicall/snapshot calldata to find the actually-billed egress

### DIFF
--- a/packages/ingest/rpcs.ts
+++ b/packages/ingest/rpcs.ts
@@ -7,7 +7,7 @@ export const latestBlocks : {
 } = {}
 
 function useArchiveNode(chainId: number, blockNumber?: bigint) {
-  if(!blockNumber) return false
+  if (blockNumber === undefined) return false
   if(!latestBlocks[chainId]) return true
   return blockNumber < latestBlocks[chainId].blockNumber - FULL_NODE_DEPTH
 }
--- a/packages/lib/blocks.spec.ts
+++ b/packages/lib/blocks.spec.ts
@@ -35,6 +35,11 @@ describe('blocks', function() {
       expect(block.number).to.equal(0n)
       expect(block.timestamp).to.equal(123n)
       expect(calls.map(call => call.blockNumber)).to.deep.equal([undefined, 0n])
+      // block 0 is the deepest block on the chain, so it must route to an
+      // archive node — guards against the falsy-bigint regression where `!0n`
+      // sent it to a full node.
+      const blockZeroCall = calls.find(call => call.blockNumber === 0n)
+      expect(blockZeroCall?.archive, 'block 0 must use an archive node').to.be.true
     } finally {
       rpcs.next = originalNext
       await cache.del('getBlock:31337:undefined')
--- a/packages/lib/blocks.spec.ts
+++ b/packages/lib/blocks.spec.ts
@@ -1,5 +1,7 @@
 import { expect } from 'chai'
-import { __estimateHeight } from './blocks'
+import { cache } from './cache'
+import { __estimateHeight, getBlock } from './blocks'
+import { rpcs } from './rpcs'
 
 describe('blocks', function() {
   it('estimates block height', async function() {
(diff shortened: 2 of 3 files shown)


murderteeth (changes requested):
## Summary

Addresses indexer `eth_getBlockByNumber` volume via longer block caching and an interpolation search in `estimateHeightManual`, and fixes `getBlock(chainId, 0n)` silently returning the head block due to a falsy-bigint check. All blocks tests pass locally (the `types.spec.ts` failure is pre-existing and unrelated). Requested changes below: correct the PR's billing premise and re-derive the cache TTLs.

## Issues

- [ ] **PR premise — Render billing.** The PR states "(Render bills inbound/total)". Per [Render's docs](https://render.com/docs/outbound-bandwidth), only outbound traffic is billed; inbound is explicitly not billed. The bytes this PR eliminates are `eth_getBlockByNumber` *response* bytes, which are inbound. This is also incongruent with the PR's own measurements: it reports outbound as "small (~tens of GB/mo)", which means the billed egress cannot be attributed to getBlock responses. Please correct the PR description. Recommend looking at outbound multicall traffic next, especially for snapshots — aggregated `eth_call` request calldata is outbound, i.e., the billed direction.
- [ ] **packages/lib/blocks.ts:24-27 — TTLs should be derived from the workload, not from immutability.** 30 days isn't derived from anything; "immutable" says a block *may* be cached forever, not that it should be. Change to: **head (no `blockNumber`) → 15 minutes; historical blocks → 1 day.** Rationale: 15m matches the indexing cycle, so every job in a cycle — apy anchors, timeseries bounds, event stride planning — sees one consistent block height per chain, and any invalidation is picked up next cycle. This also collapses the steady-state `estimateHeight` storm: with the anchor pinned, all vaults' weekly/monthly lookback timestamps (`apy.ts:172-173`) become identical and the search is deterministic, so repeated runs are pure cache hits. 1 day covers the backfill reuse window (sweeps re-probe the same day-boundary blocks for hours, then never again) at 1/30th the retention.
- [ ] **packages/lib/blocks.ts:130** — `useArchiveNode` has the same falsy-bigint bug this PR fixes in `__getBlock`: `if (!blockNumber) return false` is true for `0n`, so `getBlock(chainId, 0n)` routes to a non-archive node even though block 0 is the deepest block on the chain. Should be `blockNumber === undefined`. The new test records the `archive` flag in `calls` (blocks.spec.ts:25) but never asserts it — once fixed, asserting `archive === true` for the block-0 call would lock this in.

## Verdict

REQUEST_CHANGES

---

## How This Was Reviewed

matheus1lva:
Thanks @murderteeth — all three addressed in 91d56ae + description update:

1. **Billing premise** — corrected the description: Render bills only outbound, getBlock response bytes are inbound, so this PR cuts RPC/inbound volume not the billed egress. Flagged outbound multicall/snapshot calldata as the next target for the actually-billed direction.
2. **TTLs** — re-derived from the workload: head (no blockNumber) → 15m (one indexing cycle), historical → 1 day. Dropped the 30d "immutable" framing.
3. **`useArchiveNode` block 0** — `!blockNumber` was truthy-false for `0n`; now `blockNumber === undefined`, so block 0 routes to an archive node. Added the `archive === true` assertion on the block-0 call in `blocks.spec.ts` to lock it in.

murderteeth (approved):

PR #419, opened 2026-06-01
fix: fapy/apr-oracle for plain erc4626 (Morpho) vaults

### LINKED ISSUE #225: Fill in missing APYs

matheus1lva:
### Summary
Yearn-branded Morpho meta-vaults (plain ERC4626, no v3 interface) reported `0` for pps/apy and an empty apr-oracle in Kong. Example: `0x0963232eB842BAF53E8e517691f81745C1F228a0` (chain 1), held as a strategy by v3 allocator `0x310B7Ea7475A0B449Cfd73bE81522F1B88eFAFaa`, showed `0` while oracle.yearn.fi returned a non-zero apr.

Closes #225.

Two root causes, both fixed:
1. **pps/apy = 0**: the erc4626 hooks quoted `convertToAssets` with the *asset* decimals from `thing.defaults`. These Morpho vaults are 18-share / 6-asset, so `convertToAssets(10**6)` rounds to `0`. Now we read the vault's own immutable `decimals()` for the share-unit arg and keep asset decimals for the humanize/TVL divisor.
2. **empty apr-oracle**: plain erc4626 vaults are not classified as `yearn/3/vault`, so the apr-oracle timeseries hook never ran for them, though the oracle prices them by address. Added an erc4626 apr-oracle hook that delegates to the shared oracle resolver.

### How to review
- Start with `packages/ingest/abis/erc4626/lib.ts`. The new `ppsReadParameters()` is the single home for the share-vs-asset decimals rule. Note `decimals()` is read at `latest` (immutable, so no archive-node dependency), and a read failure warns before falling back to stored decimals (the fallback holds only when share == asset).
- `yearn/3/vault/timeseries/apr-oracle/hook.ts`: extracted `resolveOracleApr()` (oracle config + block resolution + inceptBlock guard + readApr + computeApy). The default v3 export and the new erc4626 hook both call it. `readApr` stays exported (the backfill script uses it). The v3 hook's behavior is unchanged.
- `erc4626/timeseries/apr-oracle/hook.ts` (new): emits `apr`/`apy` only (net apr/apy aren't surfaced for erc4626 vaults).
- `pps/hook.ts` + `apy/hook.ts`: both now call `ppsReadParameters(chainId, address, vault.defaults.decimals)`.

Smoke test (terminal UI after `make dev`): fanout the erc4626 timeseries hooks (terminal UI → ingest → fanout abis) for `0x0963…` on chain 1, then hit the parent allocator snapshot:
```
https://kong.yearn.fi/api/rest/snapshot/1/0x310B7Ea7475A0B449Cfd73bE81522F1B88eFAFaa
```
Strategy `0x0963…` should show non-zero pps, apy, and a populated apr-oracle (apr/apy) matching oracle.yearn.fi.

### Test plan
- [x] Automated:
  ```
  bun --filter ingest run test abis/erc4626/timeseries/apr-oracle/hook.spec.ts abis/erc4626/timeseries/pps/hook.spec.ts
  ```
  6 passing: new apr-oracle hook (pinned-block apr, no-oracle chain → `[]`, emits apr+apy); pps share≠asset decimals regression test (raw `1002039`, humanized ≈ `1.002039` at block `25211174`). Fixed avantis fixture `decimals` 18→6 (matches on-chain + prod convention).
- [x] Lint/types: `eslint` clean, `tsc --noEmit` clean.
- [ ] Manual: `make dev` → terminal UI → fanout abis → verify `0x0963…` snapshot via the allocator endpoint above.

### Post-merge recompute (#225)
These fixes only correct *new* hook runs. Already-stored history must be recomputed for affected erc4626 vaults: historical apy (`apy-bwd-delta-pps`) and historical oracle apy (`apr-oracle`).

This is timeseries data (the `output` table), not event data. Editing `evmlog_strides` and rerunning a normal fanout does **not** recompute it: the erc4626 timeseries hooks don't read evmlog, and the timeseries fanout only fills *missing* days. The wrong `apy-bwd-delta-pps` rows already exist (as `0`), so a fanout skips them and fixes only the latest point. (`apr-oracle` is a new label for erc4626 with no rows, so a fanout would backfill it on its own; the script writes it too so apy and oracle apy land in one reviewable upsert.) The native "fanout replays" command is not used.

Method: `packages/scripts/src/issue-225`. `compute.ts` recomputes into a staging table (`output_temp_fapy_oracle`), `upsert.ts` promotes it into `output` in one transaction. It reuses the live erc4626 hooks (no duplicated compute) and the shared `backfill-shared` helpers; the upsert updates `value` only on conflict, so each recomputed point matches what a normal fanout would write. `--vaults` scopes each run, `--dry-run` previews compute. Discovery queries and steps are in the script README.

Plan:
1. Merge this PR.
2. Discover affected vaults (queries in the README).
3. **Morpho first**: run compute → upsert for the Morpho subset, verify against oracle.yearn.fi, then run the remaining erc4626 vaults.
4. Refresh snapshot/timeseries caches.

### Risk / impact
- Behavior changes for vaults where `decimals()` (share) ≠ `defaults.decimals` (asset). For every standard vault (share == asset) the convertToAssets arg matches before, so nothing changes. Verified `defaults.decimals` == asset decimals by convention across `manuals.yaml` + event hooks.
- One extra `decimals()` read per pps/apy compute (cheap, at `latest`).
- No migrations, no auth/funds paths, no feature flags. Rollback = revert commit `7559ddd`.
- On a transient `decimals()` read failure the hook warns and falls back to stored decimals (could re-introduce a 0 pps for a share≠asset vault), logged with 🚨 so it's traceable rather than silent.
- Recompute (above) overwrites stored `output` rows for the targeted vaults only. It stages to `output_temp_fapy_oracle` and promotes in a single transaction; `compute.ts --dry-run` previews first; the `--vaults` list bounds the scope. No schema change. Rollback = rerun the script to re-derive the affected series.

### DIFF
--- a/packages/ingest/abis/erc4626/lib.ts
+++ b/packages/ingest/abis/erc4626/lib.ts
@@ -1,0 +1,26 @@
+import { ReadContractParameters } from 'viem'
+import { rpcs } from '../../rpcs'
+import abi from './abi'
+
+// pps must be quoted in share units (the vault's own decimals()), not the asset
+// decimals stored in thing.defaults — they differ for vaults like the 18-share /
+// 6-asset Yearn Morpho vaults, where using asset decimals rounds convertToAssets
+// to 0. decimals() is immutable, so read it at latest (avoids archive-node
+// dependency). On a read failure we warn and fall back to the stored decimals,
+// but that fallback is only correct when shares == asset, so it must be loud.
+export async function ppsReadParameters(
+  chainId: number,
+  address: `0x${string}`,
+  fallbackDecimals?: number,
+): Promise<ReadContractParameters> {
+  const shareDecimals = await rpcs.next(chainId).readContract({
+    abi, address, functionName: 'decimals'
+  }).then(d => d as bigint).catch(error => {
+    console.warn('🚨', 'erc4626 decimals() read failed, falling back to stored decimals', chainId, address, error)
+    return BigInt(fallbackDecimals ?? 0)
+  })
+
+  return {
+    abi, address, functionName: 'convertToAssets', args: [10n ** shareDecimals]
+  } as ReadContractParameters
+}
--- a/packages/ingest/abis/erc4626/timeseries/apr-oracle/hook.spec.ts
+++ b/packages/ingest/abis/erc4626/timeseries/apr-oracle/hook.spec.ts
@@ -1,0 +1,43 @@
+import { expect } from 'chai'
+import { mainnet } from 'viem/chains'
+import hook, { outputLabel } from './hook'
+import { Data } from '../../../../extract/timeseries'
+import { computeApy } from '../../../yearn/lib/apy'
+import { readApr } from '../../../yearn/3/vault/timeseries/apr-oracle/hook'
+import { getOracleConfig } from '../../../yearn/3/vault/timeseries/apr-oracle/constants'
+
+describe('abis/erc4626/timeseries/apr-oracle/hook', function() {
+  const morphoUsdt = '0x0963232eB842BAF53E8e517691f81745C1F228a0' as const
+
+  it('prices a plain erc4626 vault via the apr oracle', async function() {
+    // 0x0963 is a bare ERC4626 (no apiVersion/pricePerShare), so it lands on the
+    // erc4626 path. The oracle still prices it via getStrategyApr.
+    const oracle = getOracleConfig(mainnet.id)
+    expect(oracle).to.not.equal(undefined)
+    const apr = await readApr(mainnet.id, morphoUsdt, 25211174n, oracle!.address)
+    expect(apr).to.be.closeTo(0.11878347712984644, 1e-6)
+  })
+
+  it('returns nothing when the chain has no oracle configured', async function() {
+    const data: Data = {
+      abiPath: 'erc4626', chainId: 999999, address: morphoUsdt,
+      outputLabel, blockTime: BigInt(Math.floor(Date.now() / 1000))
+    }
+    expect(await hook(999999, morphoUsdt, data)).to.deep.equal([])
+  })
+
+  it('emits apr and apy outputs', async function() {
+    const data: Data = {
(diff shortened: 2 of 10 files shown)


murderteeth (changes requested):
after merge these changes require recompute on effected vaults for
- historical apy
- historical oracle apy

please update the PR notes with a plan and updated risk considerations. if we want to get morpho working first, it may make sense to recompute morpho first, up to you.

regarding recompute: please do not use kong's native `replay` feature. we should remove `replay` or change it's implementation. the preferred method is to edit a vault's records in the emvlog_strides table and rerun the regular abi fanout. please follow up with me to clarify.

also, note this PR will close https://github.com/yearn/kong/issues/225

matheus1lva:
> after merge these changes require recompute on effected vaults for
> 
> * historical apy
> * historical oracle apy
> 
> please update the PR notes with a plan and updated risk considerations. if we want to get morpho working first, it may make sense to recompute morpho first, up to you.
> 
> regarding recompute: please do not use kong's native `replay` feature. we should remove `replay` or change it's implementation. the preferred method is to edit a vault's records in the emvlog_strides table and rerun the regular abi fanout. please follow up with me to clarify.
> 
> also, note this PR will close #225

@murderteeth updated

murderteeth (approved):

PR #422, opened 2026-06-11
fix: consolidate erc4626 tvl hooks + zero-check no-tvl guard

matheus1lva:
## Summary

- Collapse the duplicated erc4626 `tvl` and `tvl-c` hook bodies into the shared `yearn/lib/tvl` `_process`; all six tvl hooks are now thin delegators (matching `yearn/2` + `yearn/3`). ~110 lines of duplicated `_process`/`_compute` deleted.
- `_compute`: make `apiVersion` optional so bare erc4626 vaults skip strategy delegation, and thread parsed `decimals` out instead of reading the any-typed `vault.defaults`.
- Guard no-tvl with a plain zero-check: when `priceUsd` or `totalAssets` is falsy, return `tvl: 0` (real `priceUsd` kept for the price component). Avoids polluting downstream tvl when the price source returns nothing or the vault holds nothing.

## Breaking data-format change + recompute plan

Consolidating erc4626 onto the shared `_process` changes the stored format of one timeseries component for bare erc4626 (non-yearn) vaults:

| `tvl-c` component | old erc4626 hook | new shared `_process` | change |
|---|---|---|---|
| `tvl` | `priced(...)` | `priced(...)` | none |
| `delegated` | `0n` | `0` | none |
| `totalAssets` | raw bigint (`totalAssets`) | `normalize(totalAssets, decimals)` | **÷ 10^decimals** |
| `delegatedAssets` | `0n` | `0` | none |
| `priceUsd` | `priceUsd` | `priceUsd` | none |

`totalAssets` is the only component that changes — historical rows hold the raw bigint, new rows hold the normalized value. Without a recompute the series steps by `10^decimals` at the cutover. `delegatedAssets` stays zero for these vaults (no strategy delegation), so it needs nothing.

### Scope

Only bare erc4626 vaults — the ones routed through this hook per `config/abis.yaml` (`erc4626 = true`, `yearn != true`). yearn/2 and yearn/3 already emitted normalized `totalAssets`, so they are untouched.

### Recompute: replay the `tvl-c` hook

Post-merge, run a one-shot under `packages/scripts/src/replay/` (same shape as `replay/StrategyChanged.ts`) that re-derives history from chain instead of mutating stored values:

1. Select target vaults: `thing` where `label='vault'`, `defaults->>'erc4626'='true'`, `defaults->>'yearn' IS DISTINCT FROM 'true'`.
2. Per vault, pull the existing `tvl-c` series points: `SELECT DISTINCT series_time, block_time FROM output WHERE chain_id=$1 AND address=$2 AND label='tvl-c'`.
3. For each point, call the `tvl-c` hook — `_process(chainId, address, { outputLabel: 'tvl-c', blockTime }, true)` (default export already passes `components=true`). It re-estimates block height from `blockTime`, refetches `totalAssets`/`priceUsd`, and returns normalized `Output[]`.
4. Upsert the returned rows into `output` `ON CONFLICT (chain_id, address, label, component, series_time) DO UPDATE SET value=...` — reuse `backfill-shared/upsert.ts`.
5. Batch + throttle vaults like the existing replay (chunks of ~10) to bound RPC load.

**Idempotent** — keyed upsert, so re-runs converge to the same rows; safe to resume after interruption. **Self-healing** — every point is recomputed from chain, so it also corrects any other drift in the old erc4626 history (stale price, missed blocks), not just the `totalAssets` scale.

Cost: re-fetches `totalAssets`/`priceUsd` per historical point — thousands of multicalls per vault. Heavy on RPC egress (the thing #417 was cutting), so run it off-peak, throttled, and ideally against the block-cache from #417. Acceptable as a bounded one-shot over a small vault set.

### Verify

Spot-check a known vault before/after: post-replay `tvl-c.totalAssets` should equal `tvl / priceUsd` (within rounding) at the same `series_time`, and the last historical point should line up with the first point the live indexer writes (no `10^decimals` step).

## Test plan

- `tsc --noEmit`: 0 errors
- `eslint`: 0 errors, no new warnings
- `bun --filter ingest test`: blocked locally by a pre-existing flaky chai ESM/CJS module-load race in `test.fixture.ts` (fires at fixture load, before any spec runs; unrelated to this diff)

### DIFF
--- a/packages/ingest/abis/erc4626/timeseries/tvl-c/hook.ts
+++ b/packages/ingest/abis/erc4626/timeseries/tvl-c/hook.ts
@@ -1,61 +1,9 @@
-import { z } from 'zod'
+import { Output } from 'lib/types'
 import { Data } from '../../../../extract/timeseries'
-import { EvmAddressSchema, Output, OutputSchema, Thing, ThingSchema } from 'lib/types'
-import { priced } from 'lib/math'
-import { estimateHeight, getBlock } from 'lib/blocks'
-import { first } from '../../../../db'
-import { fetchErc20PriceUsd } from '../../../../prices'
-import { rpcs } from '../../../../rpcs'
-import abi from '../../abi'
-import { extractTotalAssets } from '../../../yearn/lib/tvl'
+import _process from '../../../yearn/lib/tvl'
 
 export const outputLabel = 'tvl-c'
 
-export default async function _process(chainId: number, address: `0x${string}`, data: Data): Promise<Output[]> {
-  console.info('🧮', data.outputLabel, chainId, address, (new Date(Number(data.blockTime) * 1000)).toDateString())
-
-  let blockNumber: bigint = 0n
-  let latest: boolean = false
-  if(data.blockTime >= BigInt(Math.floor(new Date().getTime() / 1000))) {
-    latest = true;
-    ({ number: blockNumber } = await getBlock(chainId))
-  } else {
-    const estimate = await estimateHeight(chainId, data.blockTime);
-    ({ number: blockNumber } = await getBlock(chainId, estimate))
-  }
-
-  const vault = await first<Thing>(ThingSchema,
-    'SELECT * FROM thing WHERE chain_id = $1 AND address = $2 AND label = $3',
-    [chainId, address, 'vault']
-  )
-
-  if (!vault) return []
-
-  const { tvl, totalAssets, priceUsd } = await _compute(vault, blockNumber, latest)
-
-  return OutputSchema.array().parse([
-    { chainId, address, blockNumber, blockTime: data.blockTime, label: data.outputLabel, component: 'tvl', value: tvl },
-    { chainId, address, blockNumber, blockTime: data.blockTime, label: data.outputLabel, component: 'delegated', value: 0n },
-    { chainId, address, blockNumber, blockTime: data.blockTime, label: data.outputLabel, component: 'totalAssets', value: totalAssets },
-    { chainId, address, blockNumber, blockTime: data.blockTime, label: data.outputLabel, component: 'delegatedAssets', value: 0n },
-    { chainId, address, blockNumber, blockTime: data.blockTime, label: data.outputLabel, component: 'priceUsd', value: priceUsd }
-  ])
-}
-
-export async function _compute(vault: Thing, blockNumber: bigint, latest = false) {
-  const { chainId, address, defaults } = vault
-  const { asset, decimals } = z.object({
-    asset: EvmAddressSchema,
-    decimals: z.number({ coerce: true }) }
-  ).parse(defaults)
-
-  const { priceUsd } = await fetchErc20PriceUsd(chainId, asset, blockNumber, latest)
-
-  const totalAssets = await extractTotalAssets(chainId, address, blockNumber)
-
-  if(!totalAssets) return { priceUsd, tvl: 0, totalAssets, delegatedAssets: 0n }
-
-  const tvl = priced(totalAssets, decimals, priceUsd)
-
-  return { priceUsd, tvl, totalAssets }
+export default async function process(chainId: number, address: `0x${string}`, data: Data): Promise<Output[]> {
+  return _process(chainId, address, data, true)
 }
--- a/packages/ingest/abis/erc4626/timeseries/tvl/hook.ts
+++ b/packages/ingest/abis/erc4626/timeseries/tvl/hook.ts
@@ -1,58 +1,9 @@
-import { z } from 'zod'
+import { Output } from 'lib/types'
 import { Data } from '../../../../extract/timeseries'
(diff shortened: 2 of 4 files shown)


murderteeth (changes requested):
## Summary
Consolidates the duplicated erc4626 `tvl`/`tvl-c` hook bodies into the shared `yearn/lib/tvl` `_process` (matching the yearn/2 and yearn/3 delegator pattern), makes `apiVersion` optional in `_compute` so bare erc4626 vaults skip strategy delegation, threads parsed `decimals` out of `_compute` instead of re-reading any-typed `vault.defaults`, and extends the no-tvl guard to also fire when `priceUsd` is falsy.

Verified in a clean worktree at head `d34eaeb`: `tsc --noEmit` 0 errors, eslint 0 errors (35 pre-existing warnings in untouched files). Both non-yearn erc4626 manual vaults are `apiVersion 3.0.1`, so the newly-reachable delegation branch stays dormant for them. The consolidation also fixes a real bug: the old erc4626 `tvl-c` emitted raw bigints for `totalAssets`/`delegatedAssets`, which `OutputSchema.value` silently nulls (`z.number().safeParse(bigint)` fails) — the shared path emits normalized numbers.

## Issues
- [ ] **packages/ingest/abis/yearn/lib/tvl.ts:77** - The widened guard changes behavior on one narrow path: a pre-3.0.0 (v2) vault at a moment when `fetchErc20PriceUsd` returns a falsy price.

  **Before this PR** the guard was only `if (!totalAssets)`, so with `priceUsd = 0` and real `totalAssets`, execution continued past it: `delegatedAssets` was extracted for real via the withdrawalQueue RPC calls, and `tvl`/`delegatedTvl` came out 0 anyway (multiplied by price 0). Emitted components: `tvl: 0`, `delegated: 0`, `totalAssets: real`, `delegatedAssets: real`, `priceUsd: 0`.

  **After this PR** the guard is `if (!priceUsd || !totalAssets)`, so the same scenario returns early and the extraction never runs. Emitted: `tvl: 0`, `delegated: 0`, `totalAssets: real`, `delegatedAssets: 0`, `priceUsd: 0`.

  So 4 of 5 components are identical either way; the delta is the `delegatedAssets` component going from the real on-chain value to a hardcoded 0 — inconsistent within the same row set, since `totalAssets` is still emitted real. During a price outage the series shows assets fully present but delegation dropping to zero, then snapping back when the price source recovers. It doesn't affect erc4626 vaults (they never delegated), only v2 vaults during price-source gaps. Consider zeroing only the price-dependent values: extract `delegatedAssets` whenever `totalAssets` is real, and compute `tvl`/`delegatedTvl` conditionally on `priceUsd`.

## Suggestions
- **packages/ingest/abis/yearn/lib/tvl.ts:77** (non-blocking, pre-existing conflation) - `!totalAssets` treats `undefined` (multicall failure — `extractTotalAssets` already logs 🚨 and returns undefined) the same as `0n` (genuinely empty vault), recording `tvl: 0` on RPC failure as if it were real data. Since this PR's goal is keeping false zeros out of the tvl series, consider returning no outputs for the `undefined` case.

## Verdict
REQUEST_CHANGES

---

## How This Was Reviewed

matheus1lva:
Both addressed in ea40163:

- **Blocking (tvl.ts:77)** — dropped `priceUsd` from the early-return guard so it's back to `if (!totalAssets)`. `delegatedAssets` is now extracted whenever `totalAssets` is real, and `tvl`/`delegatedTvl` are computed conditionally on `priceUsd` (`priceUsd ? priced(...) : 0`). During a price-source gap a v2 vault now keeps its real `delegatedAssets` instead of snapping to 0, and the price guard still keeps NaN/garbage USD tvl out.
- **Suggestion (undefined vs 0n)** — `_process` now returns `[]` when `_compute` reports `totalAssets === undefined` (multicall failure already logs 🚨), so an RPC failure produces a gap rather than a false zero. A genuine empty vault (`0n`) still emits `tvl: 0`.

Verified: `tsc --noEmit` 0 errors, eslint clean on the touched file.

murderteeth (changes requested):
lgtm but one issue, erc4626 tvl-c data-format changes in this pr. totalAssets component change from a raw bigint to a normalized number (normalize(totalAssets, decimals)), now matching yearn/2,3. This is correct, but it means we need to recompute historical erc4626 tvl-c.totalAssets values (delegatedAssets should always be zero for non yearn vaults).

please include a plan to recompute in this PRs notes

matheus1lva:
> lgtm but one issue, erc4626 tvl-c data-format changes in this pr. totalAssets component change from a raw bigint to a normalized number (normalize(totalAssets, decimals)), now matching yearn/2,3. This is correct, but it means we need to recompute historical erc4626 tvl-c.totalAssets values (delegatedAssets should always be zero for non yearn vaults).
> 
> please include a plan to recompute in this PRs notes

updated.

murderteeth (approved):

murderteeth (approved):

PR #424, opened 2026-06-13
test: migrate test runner from mocha to vitest

matheus1lva:
### Summary
Swap the test runner from mocha + ts-node to vitest (esbuild) across `lib`, `ingest`,
and `web`. The runner changes; specs keep their chai `expect(...)` calls, so the
passing-test set is identical per package. This also fixes two pre-existing bugs:
`ingest` `bun run test` ran zero tests (mocha `--grep` filters test titles, not files),
and the `*.containers.spec.ts` exclude never matched `containers.spec.ts`, so the e2e
stack leaked into the unit run.

### Numbers
Interleaved warm samples, `bun run test` wall-clock, e2e excluded under both runners.
Full method and raw samples live in `TEST_MIGRATION_MEASUREMENTS.md`.

| package | mocha / bun | vitest | gain |
|---|---|---|---|
| ingest (56 passing) | 43.8s | 37.3s | 14.7% |
| lib (21 passing) | 12.8s | 10.4s | 18.6% |
| web (5 passing) | 0.11s | 0.53s | −0.4s |
| **aggregate** | **56.6s** | **48.3s** | **14.8% mean / 12.8% median** |

`ingest` dominates the wall-clock and runs network-bound in a keyless local env: its RPC
specs retry public nodes and fail the same way under both runners. Testcontainers
startup costs the same in both too. The win narrows to dropping ts-node's startup
type-check for esbuild. `web` regresses because `bun test` starts near-instantly for 2
trivial specs; vitest adds ~0.4s.

### How to review
- The four `vitest.config.ts` files define the setup: shared postgres+redis via
  `vitest.global.ts` (`globalSetup`), single-fork and sequential so the specs that
  write shared tables keep the old single-process isolation.
- `vitest.setup.ts` brings `rpcs` + `cache` up once and registers `chai-almost`.
- Spec diffs are mechanical: `this.timeout` / `this.skip` / `this.<state>`,
  `before`/`after`, and `Mocha.Context` ported to vitest; per-test timeouts preserved
  (`prices` 120s, `apy` 20s, `blocks` 5s) so failing-network durations stay comparable.
- `test:containers` now runs via `vitest.containers.config.ts`.

### Test plan
- [ ] `bun --filter lib test` (needs Docker for testcontainers)
- [ ] `bun --filter ingest test` (Docker; RPC-dependent specs need `.env` to pass)
- [ ] `bun --filter web test`
- [ ] e2e: `bun --filter ingest test:containers`

### Risk / impact
Test-infra only, no app or runtime code touched. The `ingest` unit run now excludes the
e2e specs, which run via `test:containers`. `web` moves off `bun test` and adds ~0.4s of
vitest startup for its 2 trivial specs. Rollback is a single revert of this commit.


murderteeth (changes requested):
## Summary

Thanks for doing the runner migration. The Vitest wiring looks generally sound, and the package scripts/checks are exercising the intended suites now.

## Dependencies

- **vitest@3.2.6**: APPROVED - dev-only test tooling, MIT, active maintenance, ESM/`sideEffects:false`, signed npm provenance; Socket reports dependency alerts resolved.

## Issues

- [ ] `TEST_MIGRATION_MEASUREMENTS.md` contains stale local decision notes that contradict this PR. It says not to open a PR and that the migration is left uncommitted, but this file is now committed in the PR. Please remove this scratch note from the PR or rewrite it as PR-appropriate benchmark documentation.

## Verification

- `bun install --frozen-lockfile` passed.
- `bun --filter web test` passed: 5 tests.
- `bun --filter web lint`, `bun --filter lib lint`, `bun --filter ingest lint` exited 0 with existing warnings.
- `bun --filter lib test` failed on Cloudflare RPC; same failure reproduced on PR base `32cc8c2`, so not introduced by this PR.
- `bun --filter ingest test` ran under Vitest: 56 passed, 3 skipped, 20 failed due unavailable/keyed RPC/API dependencies.
- `timeout 180s bun --filter ingest test:containers` reached the intended e2e stack, then timed out while the running stack hit Cloudflare RPC failures.

## Verdict

REQUEST_CHANGES

matheus1lva:
Followed up on the test failures in the Verification section — ran the suites locally with the team `.env` (the RPC/API credentials the review environment didn't have). None of the failures are migration-caused:

- **ingest:** 77 passed / 2 skipped / **0 failed** (the 20 failures were all missing RPC/API keys).
- **lib:** 26/26 passed. The `blocks` failure needs an RPC URL; the 4 `dates` failures were a non-UTC machine (CI runs UTC, so they passed in CI).
- **web:** 5/5. The `test:containers` timeout was a Cloudflare RPC stall (network), not the migration.

**Fix pushed (`1d5e1c6`):** pinned `TZ=UTC` in all four vitest configs so the date-bucketing specs are deterministic on any dev machine. `dates.spec` 5/5 and full `lib` 26/26 green locally with no shell TZ set.

**To reproduce green locally** (Docker required — testcontainers provide ephemeral postgres+redis, so no DB creds needed):
- RPC: `HTTP_ARCHIVE_<chainId>` and `HTTP_FULLNODE_<chainId>` for chains 1, 10, 100, 137, 146, 250, 8453, 42161, 80094, 747474
- Prices: `PRICE_SERVICE_URL`, `PRICE_SERVICE_API_KEY`, `YDAEMON_API`
- CRV: `CRV_GAUGE_REGISTRY_URL`, `CRV_POOLS_URL`

Without those, the network-dependent specs fail offline by design (integration tests) — identical under mocha and vitest.

murderteeth (changes requested):
## Request changes

Test-infra migration looks solid — verified `web` (5✓), `lib` (26✓), and `ingest` (e2e correctly excluded; only pre-existing RPC-dependent specs fail, runner-independent). Mechanical mocha→vitest conversions are clean: no leftover `this.timeout`/`Mocha.Context`, timeouts and skip semantics preserved.

**One change before merge:**
- [ ] **`TEST_MIGRATION_MEASUREMENTS.md`** — move to `docs/` if we want to keep it.

murderteeth (approved):

PR #425, opened 2026-06-17
refactor(web): unify list + snapshot cache refresh into a single DB pass

matheus1lva:
### Summary

Merge the two scheduled cache-refresh jobs (vault list + vault snapshots) into one. Both caches now come from a single pass over `thing` + `snapshot`, which removes the N+1 per-vault snapshot query and a duplicate scan of the same tables. Because list and snapshot data derive from one scan, the two caches no longer drift apart.

### How to review

Start with the new combined script and the query that feeds it:

- `packages/web/app/api/rest/refresh.ts` — new entrypoint that replaces both old scripts. Writes the snapshot group, the per-chain list buckets, and the `rest:list:vaults:all` bucket as independent MSETs, so one oversized or failing group cannot block the others.
- `packages/web/app/api/rest/list/db.ts` — `getVaultsWithSnapshots()` returns the curated list columns plus the raw `defaults`/`snapshot`/`hook` blobs, merged in JS as `defaults < snapshot < hook` (same shape the snapshot endpoint serves). List parse failures are kept separate from snapshot construction.

Deletions to confirm nothing else references them:

- `packages/web/app/api/rest/list/refresh.ts`
- `packages/web/app/api/rest/snapshot/refresh-snapshot.ts`
- `getVaults()` / `getVaultSnapshot()` in `snapshot/db.ts` (only the `VaultSnapshot` type remains)

Workflow + callers:

- `.github/workflows/refresh-cache.yml` replaces `snapshot-refresh.yml` and `refresh-lists.yml`.
- Container specs now invoke the new script path.

The `route.ts` change is a one-line whitespace edit; ignore it.

#### 1. Run the combined refresh against a DB

```bash
export POSTGRES_HOST=... POSTGRES_DATABASE=... POSTGRES_USER=... POSTGRES_PASSWORD=... POSTGRES_PORT=... POSTGRES_SSL=true
export REST_CACHE_REDIS_URL=redis://localhost:6379

bun packages/web/app/api/rest/refresh.ts
```

Expect a single timed run that exits `0` and logs both groups, e.g.:

```
✓ Completed: 1234 vaults cached across 6 chains, 1230 snapshots cached
```

#### 2. Confirm both cache groups landed in Redis

```bash
redis-cli GET 'rest:list:vaults:all' | head -c 120          # non-empty JSON
redis-cli KEYS 'rest:list:vaults:*'                          # one per-chain key + :all
redis-cli KEYS 'rest:snapshot:*' | wc -l                     # ~= snapshotCount above
```

#### 3. Serve the endpoints and smoke-test list + snapshot

```bash
bun --filter web dev    # next dev → http://localhost:3000
```

List (reads `rest:list:vaults:all`):

```bash
curl -s http://localhost:3000/api/rest/list/vaults | jq 'length'
curl -s 'http://localhost:3000/api/rest/list/vaults?origin=yearn' | jq '.[0]'
```

Snapshot (pick a `chainId`/`address` from the list output):

```bash
curl -s http://localhost:3000/api/rest/snapshot/1/0x310B7Ea7475A0B449Cfd73bE81522F1B88eFAFaa \
  | jq '{chainId, address, pricePerShare, apy}'
```

Expect `200` and a populated body on both. Because both caches come from the same scan, `pricePerShare` for a vault in the list response should match the same vault's snapshot response.

#### 4. Automated (container e2e — now call the new script path)

```bash
bun --filter ingest run test packages/ingest/containers.spec.ts
```

#### 5. Post-deploy smoke (prod)

```bash
curl -s https://kong.yearn.fi/api/rest/list/vaults | jq 'length'
curl -s https://kong.yearn.fi/api/rest/snapshot/1/0x310B7Ea7475A0B449Cfd73bE81522F1B88eFAFaa | jq '.pricePerShare'
```

Checklist:

- [ ] Manual: steps 1–3 (refresh writes both cache groups; list + snapshot endpoints serve consistent data)
- [ ] Automated: step 4 (container e2e)
- [ ] `tsc --noEmit` / lint clean

### Risk / impact

- Schedule change: `refresh-lists` ran every 15m and is removed; the combined `refresh-cache` runs every 30m. The list cache now refreshes every 30m instead of 15m. Confirm that cadence is acceptable.
- Uptime Kuma: the combined job now pushes to both `UPTIME_KUMA_PUSH_URL_SNAPSHOT_REFRESH` and `UPTIME_KUMA_PUSH_URL_REFRESH_LISTS`.
- No funds, auth, or migrations touched.
- Rollback: revert the workflow files and restore the two deleted scripts.


murderteeth (changes requested):
## Summary

Merges the two scheduled cache-refresh jobs into a single script (`packages/web/app/api/rest/refresh.ts`). Both the vault list and snapshot caches are now built from one pass over `thing` + `snapshot` via `getVaultsWithSnapshots()`, eliminating the N+1 per-vault snapshot query and a duplicate table scan. Workflows are consolidated (`refresh-cache.yml` replaces `snapshot-refresh.yml` + `refresh-lists.yml`), callers/docs updated, and dead code (`getVaults`/`getVaultSnapshot`/`getVaultsList`) removed.

**Verification performed:**
- ✅ `tsc --noEmit` clean for the changed `rest/*` files
- ✅ `next lint` clean on `refresh.ts`, `list/db.ts`, `snapshot/db.ts`
- ✅ No dangling references to deleted symbols/scripts (remaining `getVaults`/`getVaultSnapshot` hits are unrelated local functions in `timeseries/`, `reports/`, `monitor/`, `quality-assurance/`)
- ✅ Snapshot merge shape `{chainId, address, ...defaults, ...snapshot, ...hook}` is equivalent to the old `getVaultSnapshot`
- ✅ Snapshot cache key is safe: `getSnapshotKey` lowercases internally, matching the route's read regardless of `thing.address` casing
- ✅ `_hasSnapshot = (snapshot.chain_id IS NOT NULL)` correctly reproduces the old INNER JOIN "skip vaults without a snapshot" behavior under the new LEFT JOIN

## Issues (please address)

- [ ] **`refresh-cache.yml:31-42` — collapse to a single Uptime Kuma monitor.** The job currently pings *both* monitor URLs (`...SNAPSHOT_REFRESH` and `...REFRESH_LISTS`) on success and both on failure. Now that this is one combined job, point both the `if: success()` (status=up) and `if: failure()` (status=down) steps at the single `UPTIME_KUMA_PUSH_URL_REFRESH_VAULTS` secret (already created). Retire the two old monitors so they don't sit permanently down.

- [ ] **`list/db.ts:226` — replace `SELECT DISTINCT` with `DISTINCT ON (thing.chain_id, thing.address)`.** The current `SELECT DISTINCT` now dedups over three large jsonb blobs (`thing.defaults`, `snapshot.snapshot`, `snapshot.hook`), which is expensive and non-deterministic if two rows share a key but differ in a blob. Switch to `DISTINCT ON (chain_id, address)` for cheaper, deterministic one-row-per-vault. Note: this requires wrapping the dedup in a subquery (inner `ORDER BY chain_id, address`, outer `ORDER BY tvl DESC NULLS LAST`) so the list endpoint's TVL ordering is preserved.

- [ ] **`refresh.ts` filename — ambiguous scope.** Every other refresh script is domain-scoped (`timeseries/refresh.ts`, `reports/refresh.ts`, and the now-deleted `list/refresh.ts` / `snapshot/refresh-snapshot.ts`). A bare `rest/refresh.ts` reads as "refresh all REST caches" but only covers list + snapshot. Rename to **`rest/refresh-vaults.ts`** and update the 5 callers (`refresh-cache.yml`, both container specs, the two READMEs).

## Verdict

**REQUEST_CHANGES** — Sound refactor and the core logic is verified correct, but the Kuma monitor consolidation, `DISTINCT ON`, and the `refresh-vaults.ts` rename should land before merge.

---

## How This Was Reviewed

murderteeth (approved):

PR #428, opened 2026-06-18
perf(ingest): cut DB egress with column projection and hypertable chunk pruning

matheus1lva:
### Summary
Indexer hooks pulled more data off the database than they used: full `args` JSONB blobs on `evmlog` and unbounded hypertable scans on `output`. This trims each query to the columns and time window it needs, lowering egress and letting TimescaleDB prune chunks. Split into one commit per optimization for independent review.

### Changes
- **`getSparkline()` (`packages/ingest/db.ts`)** — 90-day `series_time` floor to prune chunks, with a full-scan fallback when the window holds fewer than 3 buckets so sparse/stale vaults keep their series.
- **evmlog projection (`tradeHandler/snapshot/hook.ts`, `yearn/2/vault/snapshot/hook.ts`)** — `SELECT signature, args` becomes `SELECT signature, args->>'<key>' AS ...` for the one or two keys each projection reads. Renamed fields (`event.seller`, `event.tokenIn`, `event.strategy`, `event.newVersion`) match the old `event.args.*` access.
- **estimated-apr (`packages/lib/estimated-apr.ts`)** — adds a `series_time` floor alongside the existing `block_time` bound. `series_time >= block_time` always holds, so the same `maxAgeDays` floor never drops a row the block_time bound keeps.
- **probe (`probe/index.ts`)** — `eventCountsCache` TTL raised from 5 minutes to 1 hour; the `count(*)` over `evmlog` is a full-table scan.

`getThing()` was dropped from this PR. It saved only tens of bytes (it still pulls the large `defaults` blob) and regressed null-`defaults` rows from a loud parse failure to a silent `undefined`.

### Impact
Measured with `EXPLAIN (ANALYZE, BUFFERS)` on the dev DB (TimescaleDB 2.13; `output` 35.3M rows / 130 chunks, `evmlog` 5.27M rows / ~7 GB).

| change | runs | measured result |
|---|---|---|
| probe TTL 5m → 1h | probe interval | full `evmlog` scan ~1.5 s / ~573k buffers (~4.5 GB IO) per run; **288 → 24 runs/day** (~264 fewer full scans/day) |
| estimated-apr `series_time` floor | per-vault APR lookup (`maxAgeDays=7`) | **chunks excluded 0 → 128/130**; buffers 6,474 → 156 (~40×); disk reads 4,740 → 37 |
| sparkline 90d floor | per vault per snapshot, ×2–3 | **chunks excluded 116/130**; buffers 2,188 → 221 (~10×); disk reads 1,608 → 0 |
| evmlog `args` projection | per v2 vault / tradeHandler snapshot | payload only (same index scan): `args` ~204 B/row → ~92 B projected = ~112 B/row off the wire (~55%) |

Relative impact: probe TTL (DB load) > estimated-apr > sparkline > evmlog projection.

### Test plan
- [ ] Manual: fanout/replay a vault; confirm timeseries (apy, pps, tvl) and snapshot hooks still produce output, and sparkline renders for active and stale vaults.
- [ ] Manual: verify estimated-apr rows return for vaults inside and outside the `maxAgeDays` window.
- [ ] Automated: `bun --filter ingest test`, `bun --filter lib test`

### Risk / impact
The 90-day sparkline floor and the estimated-apr `series_time` floor narrow query windows. The sparkline fallback covers sparse history and `series_time >= block_time` keeps the apr floor safe; both are the areas to sanity-check. No migrations, auth, or funds touched. Rollback is reverting the branch.

### DIFF
--- a/packages/ingest/abis/yearn/2/tradeHandler/snapshot/hook.ts
+++ b/packages/ingest/abis/yearn/2/tradeHandler/snapshot/hook.ts
@@ -16,7 +16,7 @@ async function project(chainId: number, tradeHandler: `0x${string}`) {
   ]
 
   const events = await db.query(`
-  SELECT signature, args
+  SELECT signature, args->>'seller' AS seller, args->>'tokenIn' AS "tokenIn"
   FROM evmlog
   WHERE chain_id = $1 AND address = $2 AND signature = ANY($3)
   ORDER BY block_number ASC, log_index ASC`,
@@ -26,13 +26,13 @@ async function project(chainId: number, tradeHandler: `0x${string}`) {
   const result: Partial<Tradeable>[] = []
 
   for (const event of events.rows) {
-    const index = result.findIndex(r => r.strategy === zhexstring.parse(event.args.seller) && r.token === zhexstring.parse(event.args.tokenIn))
+    const index = result.findIndex(r => r.strategy === zhexstring.parse(event.seller) && r.token === zhexstring.parse(event.tokenIn))
     switch (event.signature) {
     case topics[0]:
       if (index > -1) break
       result.push({
-        strategy: zhexstring.parse(event.args.seller),
-        token: zhexstring.parse(event.args.tokenIn)
+        strategy: zhexstring.parse(event.seller),
+        token: zhexstring.parse(event.tokenIn)
       })
       break
     case topics[1]:
--- a/packages/ingest/abis/yearn/2/vault/snapshot/hook.ts
+++ b/packages/ingest/abis/yearn/2/vault/snapshot/hook.ts
@@ -136,7 +136,7 @@ export async function projectStrategies(chainId: number, vault: `0x${string}`, b
   ]
 
   const events = await db.query(`
-  SELECT signature, args
+  SELECT signature, args->>'strategy' AS strategy, args->>'newVersion' AS "newVersion"
   FROM evmlog
   WHERE chain_id = $1 AND address = $2 AND signature = ANY($3) AND (block_number <= $4 OR $4 IS NULL)
   ORDER BY block_number ASC, log_index ASC`,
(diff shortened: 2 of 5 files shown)


murderteeth (changes requested):
## Summary
This PR contains several distinct optimization mechanisms bundled into one commit. The PR body explains the intent, but the implementation is difficult to evaluate because the changes are not separated by risk profile and there are no impact estimates or measurements.

## Issues

- [ ] **packages/ingest/db.ts:115** - Please roll back `getThing()` unless you can show that it makes a material runtime difference.

  This helper replaces `SELECT * FROM thing` with `SELECT defaults`, then reconstructs `chainId`, `address`, and `label` client-side. Since `thing` only appears to contain `chain_id`, `address`, `label`, and `defaults`, this likely saves only tens of bytes per call plus minimal camel-case conversion work. That is not obviously material compared with the larger JSONB and hypertable changes.

  It also changes behavior for rows where `defaults` is null: the previous `ThingSchema` parse would fail loudly, while `getThing()` returns `undefined` and makes callers silently short-circuit. If this optimization is worth keeping, please provide evidence: call frequency, before/after query payload size, timing, or production metric impact.

- [ ] **Commit structure** - Please split this into separate commits, one per optimization.

  This PR currently has one commit, `fix: last findings for egree reduction`, but it contains at least five distinct changes:

  1. `thing` lookup projection via `getThing()`
  2. `getSparkline()` 90-day `series_time` floor plus fallback
  3. `evmlog` JSONB key projection
  4. `estimated-apr` `series_time` pruning
  5. probe `eventCountsCache` TTL change

  These changes have different risk profiles and should be reviewable independently. If `getThing()` is removed, the remaining optimizations should still be split into focused commits.

- [ ] **PR evidence** - Please add an estimate of the effect of each optimization.

  The estimate should account for how each change lands at runtime. For example: whether the function runs per vault, per fanout, per snapshot, per request, or on a probe interval; expected rows scanned; expected payload reduction; and before/after `EXPLAIN` or timing where applicable.

  The current PR states the intended benefits, but it does not provide enough evidence to judge which changes materially reduce egress or database load.

## Suggestions
The `evmlog` projection and `series_time` pruning changes seem much more likely to matter than `getThing()`, but they still need per-change estimates so reviewers can validate the tradeoff.

## Verdict
REQUEST_CHANGES

---

## How This Was Reviewed
This review was conducted using the review-pr workflow via ship-spec.

murderteeth (approved):
Prior concerns addressed: getThing dropped, split into one commit per optimization, and per-change impact evidence added. Lint clean, changes correct and consistent with the existing series_time chunk-pruning pattern. Thank you. LGTM.

PR #431, opened 2026-06-21
perf(web): collapse timeseries refresh N+1 into single scan

matheus1lva:
## What

Collapse the two timeseries cache-refresh jobs from per-vault×label query loops into a handful of queries that build the `{"value":[…]}` cache envelope directly in SQL (`jsonb_agg`), instead of returning flat rows and reshaping in JS.

- **Hourly** (`refresh.ts`, cron `0 * * * *`, 2-day window): one chunk-pruned scan **per label** across all vaults. Vault-things only, via an `EXISTS` semi-join.
- **Daily** (`refresh-historical.ts`, cron `0 0 * * *`, full history): one indexed query **per vault** for all 4 labels at once (`GROUP BY label`), looped over vaults in batches.

Both still write a key for **every** vault×label each run (empty `{"value": []}` when the vault has no data in scope), so a key that loses its data is cleared — no TTL needed.

## Why

Per-row, the old queries re-sent `chain_id` + the 42-char checksummed `address` + `label` + a constant `'1 day'` `period` on **every** daily bucket. Building the series as one `jsonb` payload per vault drops that repetition, which is the bulk of the egress (and Neon bills egress).

## Measured (prod scale: N=2,174 vaults, 11 chains)

Run against the read replica from `packages/web/.env`. Wall-clock is from a laptop (WAN RTT, `pool=4`) — **same conditions on both sides**, so the comparison is fair; in-region CI is far faster and the absolute gap shrinks. The RTT-independent wins are **query count** and **egress bytes**. Daily figures are extrapolated ×29.8 from a stratified 73-vault sample (query counts are exact).

| | Hourly (recent) | Daily (full) |
|---|---|---|
| queries/run | **8,696 → 5** (4 scans + vault list) | **8,696 → 2,174** (1 indexed query/vault) |
| **egress/run** | **21.9 → 9.8 MB (−55%)** | **~4.5 → ~1.9 GB (−58%)** |
| rows over the wire | 129,468 → 7,890 | flat rows → ≤4 rows/vault |
| wall-clock (laptop) | 394 s → 7.9 s | ~36 min → ~20 min (est.) |
| peak RSS | 113 → 82 MB | 167 → 145 MB |

Combined egress ≈ **148 → 63 GB/month (~85 GB/mo saved)**, dominated by the daily job.

## Risk analysis

- **Stale `latest` keys — none.** Every run writes a key for every vault×label, defaulting to `{"value": []}` when there's no data in window, so a vault that drops out of the 2-day window has its key emptied (not left to override historical). This matches the original pre-refactor behavior; no TTL is introduced.
- **Blast radius.** No single giant query: hourly `cacheMSet`s per label (4 flushes), daily per 10-vault batch. A failure aborts the run but already-flushed keys keep their fresh values — same partial-progress profile as the old loop. Every query is either chunk-pruned (2-day hourly) or single-vault indexed (daily), so none is large enough to be cancelled by the replica's `conflict with recovery` (which a whole-table/whole-chain scan *does* hit — that's why daily stays per-vault).
- **`EXISTS` semi-join.** Planner hashes the 2,174 vault rows from `thing` (via `thing_idx_label`, ~200 kB, ~3.5 ms) and hash-joins the chunk-pruned `output` scan. `thing.address` and `output.address` are both checksummed (verified), so it matches on raw equality — no `lower()` (which would add a per-row function eval over the 129k-row scan for zero gain) and no functional index needed.
- **Rollback.** Code-only — no schema/migration and no `cache.ts` change. Revert restores the per-vault loop; the next cron tick rewrites every key.

## Behavior / contract

- **Payloads byte-identical.** Verified old-vs-new SQL on real vaults across all 4 labels: `value` stays a numeric **string** (`numeric::text`, as node-pg returned it before), `time` stays epoch-**seconds** as a JSON number. Only JSON whitespace and intra-timestamp key order differ — the read route `JSON.parse`s and re-sorts by `time`, so responses are identical.
- Redis key formats (`redis.ts`), the read endpoint, and `cache.ts` are untouched. No new dependencies.
- Removed the now-dead `period` column and the per-vault-per-label fan-out.

## Tests

`tsc --noEmit` and `next lint` clean. Correctness validated by deep-comparing old vs new query output against the live DB (full-path 5/5 vaults, recent-path 8/8, 0 mismatches). Note: this rewrite drops the prior head's testcontainers integration spec; happy to port a `refresh.spec.ts` for the new shape if we want it back before merge.

### DIFF
--- a/packages/web/app/api/rest/timeseries/db.ts
+++ b/packages/web/app/api/rest/timeseries/db.ts
@@ -16,6 +16,12 @@ export type TimeseriesRow = {
   time: bigint
 }
 
+export type SeriesPayloadRow = {
+  chainId: number
+  address: string
+  payload: string
+}
+
 export async function getVaults(): Promise<VaultRow[]> {
   const result = await db.query(`
     SELECT DISTINCT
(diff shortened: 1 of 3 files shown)


murderteeth (changes requested):
**Requesting changes before merge — two things missing from the PR description:**

### 1. Quantitative estimate per job

The PR claims "one scan instead of N*4" but never sizes the win. I want individual before/after numbers for each, since they run on different cadences and different data shapes:

- **Hourly refresh** (`refresh.ts`, cron `0 * * * *`) — uses `getAllRecentTimeseries` (2-day window). State current vault count `N`, old query count (`N×4`), and the actual gain: query count, wall-clock, and compute/billed-time before vs after on prod-sized data. This is the job that runs 24×/day, so its per-run cost matters most.
- **Daily historical** (`refresh-historical.ts`, cron `0 0 * * *`) — uses `getAllFullTimeseries` (full table, no window). Same before/after. This one also now `GROUP BY`s the entire `output` table and materializes every daily-bucketed row in node memory in a single pass — give the measured peak rows/memory at current scale so the new ceiling is a number, not the "ponytail" hand-wave in the code comment.

Local test timings (27ms / 2.8ms) are on a 2-vault fixture and don't count.

### 2. Risk analysis of merging

No PR should merge without one. At minimum address:

- **Stale `latest` keys.** The old code always wrote `{value:[]}` for every vault×label; the new code skips empty results. Keys have no TTL, so an hourly-refresh key that previously had data and now has zero rows in the 2-day window is never overwritten/emptied. Argue why the read-path merge makes this safe (or fix it) — the current "missing key reads as `[]`" note only covers never-written keys, not this case.
- **Single-scan blast radius.** Both jobs now depend on one large query succeeding. What's the failure/timeout behavior at scale vs. the old per-vault loop that failed partially?
- **`EXISTS` semi-join.** `lower(t.address) = lower(o.address)` can't use a plain index on `thing.address`; confirm the planner choice (hash semi-join) holds at current row counts, or add a functional index.
- **Rollback.** Confirm revert is clean (it should be — Redis keys are rewritten by the prior impl on the next cron tick).

---

## How This Was Reviewed

murderteeth (changes requested):
- Can I get these performance estimates in the form of compute hours? that's how neon bills us
- It seems clear we get some kind of win on the hourly job. But do we gain enough on the historical job that it's worth making more complicated? We can also just run the historical refresh less often, maybe just once a week, for an easier win

matheus1lva:
> **Requesting changes before merge — two things missing from the PR description:**
> 
> ### 1. Quantitative estimate per job
> 
> The PR claims "one scan instead of N*4" but never sizes the win. I want individual before/after numbers for each, since they run on different cadences and different data shapes:
> 
> - **Hourly refresh** (`refresh.ts`, cron `0 * * * *`) — uses `getAllRecentTimeseries` (2-day window). State current vault count `N`, old query count (`N×4`), and the actual gain: query count, wall-clock, and compute/billed-time before vs after on prod-sized data. This is the job that runs 24×/day, so its per-run cost matters most.
> - **Daily historical** (`refresh-historical.ts`, cron `0 0 * * *`) — uses `getAllFullTimeseries` (full table, no window). Same before/after. This one also now `GROUP BY`s the entire `output` table and materializes every daily-bucketed row in node memory in a single pass — give the measured peak rows/memory at current scale so the new ceiling is a number, not the "ponytail" hand-wave in the code comment.
> 
> Local test timings (27ms / 2.8ms) are on a 2-vault fixture and don't count.
> 
> ### 2. Risk analysis of merging
> 
> No PR should merge without one. At minimum address:
> 
> - **Stale `latest` keys.** The old code always wrote `{value:[]}` for every vault×label; the new code skips empty results. Keys have no TTL, so an hourly-refresh key that previously had data and now has zero rows in the 2-day window is never overwritten/emptied. Argue why the read-path merge makes this safe (or fix it) — the current "missing key reads as `[]`" note only covers never-written keys, not this case.
> - **Single-scan blast radius.** Both jobs now depend on one large query succeeding. What's the failure/timeout behavior at scale vs. the old per-vault loop that failed partially?
> - **`EXISTS` semi-join.** `lower(t.address) = lower(o.address)` can't use a plain index on `thing.address`; confirm the planner choice (hash semi-join) holds at current row counts, or add a functional index.
> - **Rollback.** Confirm revert is clean (it should be — Redis keys are rewritten by the prior impl on the next cron tick).
> 
> ---
> 
> ## How This Was Reviewed
> This review was conducted using the [review-pr skill](https://github.com/yearn/webops/blob/main/skills/review-pr/SKILL.md).
> 

Yep, I was doing that and left for some chores and Brazil World Cup game 😬

I stumbled on the recovery mode error often so I'm trying to build more heuristics to Bench correctly - I'll let you know when I add it!

PR #435, opened 2026-06-27
fix(security): bound public GraphQL queries and harden webhook ingest

matheus1lva:
### Summary

A security scan flagged six issues across the public API and webhook ingest. Anonymous callers could force full-table scans and uncapped Timescale aggregations through the `prices`, `timeseries`, and `tvls` GraphQL resolvers, and the webhook extractor accepted out-of-scope receiver output, read unbounded response bodies, and followed receiver-controlled redirects. This branch bounds each path.

- **Public GraphQL (findings 1-3):** `prices` now requires an `address` (timestamp alone is a near-full table scan); `timeseries` and `tvls` clamp `limit` to 1000 and reject any `period` outside a coarse allowlist. A 50k-row backstop caps `prices` responses.
- **Webhook output scope (findings 4-5):** receiver outputs whose `chainId`/`address` fall outside the fanout request are dropped before the load queue, which also bounds total groups to the requested vault set.
- **Webhook fetch (findings 4, 6):** a 30s timeout and a 10 MiB capped reader stop a receiver from stalling workers or exhausting memory; `redirect: 'manual'` plus a public-https-only subscription URL check block SSRF pivots.

### How to review

Start with `packages/web/app/api/gql/resolvers/guards.ts` (the shared `clampLimit`/`resolvePeriod`) and how `prices.ts`/`timeseries.ts`/`tvls.ts` apply it. Then `packages/ingest/extract/webhook.ts`: `selectValidOutputs` (scope filter, extracted as a pure function) and `readJsonCapped` plus the hardened `fetchResponse`. Finish with `isPublicHttpsUrl` in `packages/lib/subscriptions.ts`. The `.spec.ts` files alongside each show the intended boundaries.

### Test plan

- [ ] Automated: `bun --filter web test` (guards spec, no containers) and `bun --filter ingest test` / `bun --filter lib test` (webhook + subscriptions specs; these spin up postgres/redis testcontainers, so Docker must be running).
- [ ] Manual: smoke the public GraphQL endpoint. `prices` with no `address` should error, `timeseries`/`tvls` with `period: "1 second"` or `limit: 999999` should reject/clamp, and a normal `prices(chainId, address)` / `tvls(chainId, address)` query should still return data.

### Risk / impact

Behavioral changes a client could hit:
- `prices` rejects calls without an `address` (e.g. bare `{ prices }` or `prices(chainId: 1)`). Documented usage always passes `address`.
- `timeseries`/`tvls` reject sub-day or arbitrary `period` values and clamp `limit` above 1000.
- Webhook receivers that return a 3xx, take over 30s, or send over 10 MiB now fail that job.
- Subscription URLs must be public https. A private or http URL in `config/subscriptions.yaml` fails config load at startup, and all committed URLs already pass.

No DB migrations, no auth changes, no funds-handling code. Rollback is a branch revert.

### DIFF
--- a/packages/ingest/extract/webhook.spec.ts
+++ b/packages/ingest/extract/webhook.spec.ts
@@ -1,0 +1,73 @@
+import { strict as assert } from 'node:assert'
+import { selectValidOutputs, MAX_OUTPUTS_PER_VAULT } from './webhook'
+import type { Data } from './webhook'
+import type { Output } from 'lib/types'
+import type { WebhookSubscription } from 'lib/subscriptions'
+
+const VAULT_A = '0x1111111111111111111111111111111111111111'
+const VAULT_B = '0x2222222222222222222222222222222222222222'
+const OUT_OF_SCOPE = '0x3333333333333333333333333333333333333333'
+
+const subscription: WebhookSubscription = {
+  id: 'S_TEST',
+  url: 'https://example.com/webhook',
+  abiPath: 'yearn/2/vault',
+  type: 'timeseries',
+  labels: ['apr']
+}
+
+function data(vaults: string[], chainId = 1): Data {
+  return {
+    abiPath: 'yearn/2/vault',
+    chainId,
+    blockNumber: 1n,
+    blockTime: 1n,
+    subscription,
+    vaults: vaults as `0x${string}`[]
+  }
+}
+
+function output(address: string, label = 'apr', chainId = 1): Output {
+  return {
+    chainId,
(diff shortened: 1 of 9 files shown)


murderteeth (changes requested):
please update the commit message on 5af2b32a32afc26b6193dadfb0ad6d522bb18715

matheus1lva:
@murderteeth updated.

murderteeth (approved):
lg ty

PR #440, opened 2026-07-15
fix(ingest): remove webhook response filtering

matheus1lva:
## Summary

Restore yvUSD strategy APR ingestion by removing the webhook response filtering introduced in #435.

- Remove selectValidOutputs and queue schema-valid webhook outputs directly.
- Remove requested-address filtering and composition authorization machinery.
- Retain response-size, timeout, redirect, and public-HTTPS URL protections.
- Add extractor-level regression coverage proving a webhook triggered by one vault can publish an output for another address without composition lookup.

## Root cause

The requested-address gate introduced in #435 silently rejected legitimate strategy APR outputs because the triggering vault and returned strategy address differ.

Configured webhook services are trusted data producers, so the extractor now passes OutputSchema-valid responses directly to the load queue.

## Validation

- packages/ingest: bun run test extract/webhook.spec.ts (4 passed)
- packages/ingest: bunx eslint extract/webhook.spec.ts
- git diff --check passed

### DIFF
--- a/packages/ingest/extract/webhook.spec.ts
+++ b/packages/ingest/extract/webhook.spec.ts
@@ -1,11 +1,11 @@
 import { strict as assert } from 'node:assert'
-import { selectValidOutputs, readJsonCapped, MAX_OUTPUTS_PER_VAULT } from './webhook'
+import { mq } from 'lib'
+import { readJsonCapped, WebhookExtractor } from './webhook'
 import type { Data } from './webhook'
 import type { Output } from 'lib/types'
 import type { WebhookSubscription } from 'lib/subscriptions'
 
 const VAULT_A = '0x1111111111111111111111111111111111111111'
-const VAULT_B = '0x2222222222222222222222222222222222222222'
 const OUT_OF_SCOPE = '0x3333333333333333333333333333333333333333'
 
 const subscription: WebhookSubscription = {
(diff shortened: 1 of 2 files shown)


murderteeth (changes requested):
## Summary

The PR correctly identifies the yvUSD regression introduced by #435, but the proposed `requested-or-composition` authorization regime adds unnecessary complexity around a trust boundary that does not provide meaningful data integrity.

## Issues

- [ ] **packages/ingest/helpers/webhook.ts** - Remove address-scope authorization instead of adding composition scope.

  Please remove the requested-address gate introduced in #435 and remove the new `outputScope: requested-or-composition` feature from this PR.

  Configured webhook services are trusted data producers. Restricting returned addresses does not provide meaningful data-integrity protection: a compromised or incorrect service can already poison authorized addresses through false values, timestamps, block metadata, or components. The gate only blocks one narrow output pattern, while it has already silently rejected legitimate strategy APR data since July 1.

  Adding a database-backed composition authorization system makes the webhook contract substantially more complex without fixing that underlying trust model.

  Please retain the useful structural and resource protections:

  - requested-chain validation;
  - configured-label validation;
  - per-address and total output/group caps;
  - response-size and timeout limits;
  - redirect and URL restrictions.

  The intended contract should be: a configured webhook may publish its configured labels for addresses on the requested chain, within bounded resource limits.

  Please also add regression coverage showing that a webhook can return outputs for an address other than the triggering vault, without requiring composition lookup.

## Verdict

REQUEST_CHANGES

---

## How This Was Reviewed

matheus1lva:
> ## Summary
> The PR correctly identifies the yvUSD regression introduced by #435, but the proposed `requested-or-composition` authorization regime adds unnecessary complexity around a trust boundary that does not provide meaningful data integrity.
> 
> ## Issues
> * [ ]  **packages/ingest/helpers/webhook.ts** - Remove address-scope authorization instead of adding composition scope.
>   Please remove the requested-address gate introduced in #435 and remove the new `outputScope: requested-or-composition` feature from this PR.
>   Configured webhook services are trusted data producers. Restricting returned addresses does not provide meaningful data-integrity protection: a compromised or incorrect service can already poison authorized addresses through false values, timestamps, block metadata, or components. The gate only blocks one narrow output pattern, while it has already silently rejected legitimate strategy APR data since July 1.
>   Adding a database-backed composition authorization system makes the webhook contract substantially more complex without fixing that underlying trust model.
>   Please retain the useful structural and resource protections:
>   
>   * requested-chain validation;
>   * configured-label validation;
>   * per-address and total output/group caps;
>   * response-size and timeout limits;
>   * redirect and URL restrictions.
>   
>   The intended contract should be: a configured webhook may publish its configured labels for addresses on the requested chain, within bounded resource limits.
>   Please also add regression coverage showing that a webhook can return outputs for an address other than the triggering vault, without requiring composition lookup.
> 
> ## Verdict
> REQUEST_CHANGES
> 
> ## How This Was Reviewed
> This review was conducted using the [review-pr skill](https://github.com/yearn/webops/blob/main/skills/review-pr/SKILL.md).

Reverted, updated and test added.

murderteeth (changes requested):
This revision does not address the previous review correctly.

You are expected to review your changes against the requested requirements before presenting them for review. Reviewers should receive your best completed work at all times.

matheus1lva:
I manually dropped by hand all that was added on the gate and apparently i missed the limit output. Though today was not a good day for me, im gonna retry tomorrow.

murderteeth (approved):

PR #442, opened 2026-07-18
feat(prices): gate price source on USE_PRICE_SERVICE and stop serving table-backed prices

matheus1lva:
### Summary

`USE_PRICE_SERVICE` is a **hard switch** for the indexer price path:

- **`false` (default):** legacy Postgres `price` table path keeps its fallback order (yDaemon → DB → lens → yprice → service). Two incidental fixes ride along in this path: `load.price` no longer enqueues on an `undefined` yDaemon result, and the last-resort `fetchPriceServiceUsd` fallback picks up the corrected chain names (chain 100 → `gnosis`, added sonic/berachain/katana) and drops the `source=defillama` filter.
- **`true`:** yearn-prices is the **only** source — no table read/write, no yDaemon/lens/yprice fallbacks. Missing service data → `priceSource: 'na'`.

GraphQL/REST read layer no longer serves table-backed per-block prices (those were yDaemon spot snapshots at ingest, not historical). `prices` resolver returns `[]`; `tvls` keeps `priceUsd`/`priceSource` as null/`'na'` for response-shape compatibility.

### How to review

Start in `packages/ingest/prices.ts`:

- `usePriceService()` → `__fetchErc20PriceUsd` branches hard.
- Service mode: `fetchPriceServiceUsd` only; `maybePersistPrice` / waveydb gates skip table writes.
- Read side: `resolvers/prices.ts` empty; `resolvers/tvls.ts` drops the `price` join.
- Incidental: REST list adds `thing.address` as a final name COALESCE fallback (`rest/list/db.ts`), so unnamed vaults render their address instead of `null`.

### Test plan

- [ ] Manual: `USE_PRICE_SERVICE=false` — ingest + TVL behave as before.
- [ ] Manual: `USE_PRICE_SERVICE=true` — historical from price service only; `prices` → `[]`; `tvls` → `priceUsd: null`.
- [ ] Automated: `bun --filter ingest test`, `bun --filter web test`.

### Risk / impact

Default remains off. Service mode needs process restart. GraphQL `prices` empty for everyone when this lands on the web package — point consumers at prices.yearn.dev. Field kept for compatibility; full removal is a follow-up.


murderteeth (dismissed):
The hard-switch shape is right: one `usePriceService()` gate, legacy path untouched by default, and the `tvls` CTE collapse is a genuine simplification.

### Issues

- [ ] **[HIGH] Misconfigured service mode writes permanent zeros** — `packages/ingest/prices.ts:43`
-- With `USE_PRICE_SERVICE=true` and no `PRICE_SERVICE_API_KEY`, an unmapped chain, or a 401/429, every token silently becomes `priceUsd: 0 / 'na'`, and `abis/yearn/lib/tvl.ts:88` turns that into a `tvl=0` row that outlives the outage. Nothing at startup or runtime distinguishes the misconfiguration from a genuine absence of price data.
-- Provenance: guard existed on this branch and was removed — `assertPriceSourceConfig()` added in `9fa865b`, dropped in `0ada81b`.
-- Change: restore the startup abort when `usePriceService() && !PRICE_SERVICE_API_KEY`, and log the miss reason (no key / unmapped chain / HTTP status / empty coins) in the service branch — the legacy path already warns at `prices.ts:84`.
-- Keep: a correctly-configured service with genuinely no price still returns `'na'` rather than throwing.
-- Done when: startup with service mode and no key fails loudly, and a non-200 emits a warn carrying its status code.

- [ ] **[HIGH] Service mode fetches once per block instead of once per day** — `packages/ingest/prices.ts:33`
-- Every lookup is keyed `(chain, token, blockNumber)` for 30s, and service mode also skips the DB read that used to absorb repeats — so one token touched at 40 blocks in a day is 40 calls to prices.yearn.dev, all returning the same day-bucketed value.
-- Provenance: regression within the branch — day-keyed cache added in `6c6a426`, deleted in `40560ad` along with `prices.service-cache.spec.ts`.
-- Change: restore the past-day branch — cache successful `priceservice` results under a `(chain, token, UTC day)` key with a long TTL, keep the 30s block key for the current day only, and don't cache `na` under the day key.
-- Keep: `USE_PRICE_SERVICE=false` keeps its existing 30s block-keyed behaviour.
-- Done when: N distinct blocks for one token within one UTC day produce exactly one service fetch, with the deleted spec back and green.

- [ ] **[MEDIUM] New spec silently disables 18 unrelated tests** — `packages/ingest/prices.service.spec.ts:21`
-- The top-level `vi.resetModules()` re-evaluates `lib/global.ts`, whose `BigInt.prototype.toJSON` define is non-configurable, so `extract/evmlogs.spec.ts` and `abis/yearn/3/vault/timeseries/apr-oracle/hook.spec.ts` fail to collect at all. Confirmed locally: 84 passing with this file, 90 without it.
-- Provenance: introduced in `40560ad` (the file is new in that commit).
-- Change: don't reset the shared registry from a spec — move `usePriceService()` into `packages/ingest/helpers/env.ts` (no `lib` cycle, so no barrel mock is needed), or make `lib/global.ts` idempotent.
-- Keep: both service-mode assertions in this file.
-- Done when: `bunx vitest run` in `packages/ingest` reports no `Cannot redefine property: toJSON`, and only the 3 pre-existing network-dependent files fail.

---

## How This Was Reviewed
5 review lenses, each finding independently verified by codex. 3 candidate findings were refuted and dropped.

murderteeth (commented):
Superseded by the review below — posted with the wrong state, same content.

murderteeth (dismissed):
All three prior findings are fixed: the startup gate is back with per-reason miss logging, past-day results are day-keyed at 24h with `na` correctly excluded, and the ingest suite collects cleanly. Two problems in the fix itself.

### Issues

- [ ] **[HIGH] Service-outage misses are no longer deduplicated** — `packages/ingest/prices.ts:63`
-- A past-day lookup that finds no price is now cached nowhere, so every repeat of the same block re-issues the HTTP request and logs another warning. Unpriceable tokens are routine during historical replay, which re-creates the request amplification the day cache was added to stop — in the one case where the service is already struggling.
-- Provenance: regression from `40560ad`, where every service-mode lookup was dampened by the 30s block-keyed `cache.wrap`; introduced in `71327f9`.
-- Change: on a day-key miss, fall through to the `cache.wrap` at line 67 with `knownBlockTime` threaded in, instead of calling `__fetchErc20PriceUsd` directly; promote only a successful result to the day key.
-- Keep: `na` still never written under the day key; the existing `fetches once for many blocks in the same utc day` spec still passes.
-- Done when: two back-to-back past-day lookups of the same `(chain, token, block)` with `fetch` stubbed `{ ok: false }` issue exactly one fetch, asserted in `prices.service-cache.mock.spec.ts`.

- [ ] **[MEDIUM] Price lookups can throw during ingest startup** — `packages/ingest/prices.ts:57`
-- Jobs draining the queue before the Redis cache finishes connecting now crash with `TypeError: Cannot read properties of undefined (reading 'get')` instead of quietly fetching uncached. `index.ts:82` starts `cache.up()` and the worker pools concurrently, so the window is real on every restart with a backlog.
-- Provenance: introduced in `71327f9` — first code on the price path to use `cache.get`/`cache.set` rather than `cache.wrap`, which is the only accessor in `packages/lib/cache.ts:26` with an uninitialised fallback.
-- Change: give `__Cache.get`/`__Cache.set` the same guard `wrap` has (`get` -> `undefined`, `set` -> no-op), or guard the two calls here so an un-upped cache falls through to the direct fetch.
-- Keep: once the cache is up, a day-key hit still short-circuits the fetch and a `priceservice` result is still written with `PAST_DAY_CACHE_TTL_MS`.
-- Done when: `fetchErc20PriceUsd(chainId, token, pastDayBlock)` in service mode without `cache.up()` returns a price instead of rejecting.

---

## How This Was Reviewed
5 review lenses, each finding independently verified by codex. 3 candidate findings were refuted and dropped.

murderteeth (approved):

PR #443, opened 2026-07-20
fix(ingest): default estimated apr components so composition parse survives netAPR/netAPY-only rows

matheus1lva:
### Summary

Since 07-17, two Katana vaults (`0x80c34B…`, `0x9A6bd7…`) stopped rewriting snapshots: every `extract.snapshot` threw a ZodError, so REST kept stale `strategies[].performance` with no `estimated` block (downstream missing `strategyRewardsAPR` / `katRewardsAPR`).

Cause: `fetchStrategyPerformance` only set `components` when a non-netAPR/netAPY component appeared. Strategies whose latest `katana-estimated-apr` rows carried only `netAPR`/`netAPY` produced `{type, apr, apy}` without `components`. `CompositionSchema` requires that key, so `extractComposition` failed and the whole vault snapshot was never loaded.

Fix in `fetchStrategyPerformance`: initialize `estimated` as `{ type, components: {} }`. Also report composition parse failures to Sentry (`hook: vault.snapshot.extractComposition`) before rethrow.

Other small fixes that landed while making the e2e real:
- `projectStrategies` no longer returns `[]` when there are no `StrategyChanged` events; it still merges `get_default_queue` (needed when history is capped).
- Webhook `selectValidOutputs` groups with a plain `Map` instead of `Map.groupBy` (missing on some runtimes).
- Container test env forwards `WEBHOOK_SECRET_*`.
- Katana + yvUSD container specs delete orphan vault things (from `StrategyChanged`) before `refresh-vaults` so Zod name checks do not fail.

### How to review

1. `packages/ingest/abis/yearn/3/vault/snapshot/hook.ts` — `components: {}` init, Sentry wrap on parse, empty-events `projectStrategies` path.
2. Unit case in `hook.spec.ts` for netAPR/netAPY-only rows.
3. `katana-estimated-apr.containers.spec.ts` — real Katana parent/strategy addresses, seeded estimated-apr (live `S_KATANA_APR` currently returns `[]` for these vaults), sibling freeze shape, gate on composition `katRewardsAPR`.
4. Skip the vitest/tsconfig chore unless you care about parallel container forks.

### Test plan

- [ ] Manual: after deploy, `fanout abis` on Katana; confirm `extract-747474` no longer piles ZodErrors for `0x80c34B…` / `0x9A6bd7…`, and `/api/rest/snapshot/747474/0x80c34bd3a3569e126e7055831036aa7b212cb159` serves fresh `strategies[].performance.estimated` with rewards components where the APR service has them.
- [x] Automated: `vitest run abis/yearn/3/vault/snapshot/hook.spec.ts` (includes netAPR/netAPY-only case).
- [x] Automated e2e: `bun run test:containers packages/ingest/katana-estimated-apr.containers.spec.ts` (2/2 pass; asserts parent composition surfaces strategy `katRewardsAPR`).

### Risk / impact

Low. Ingest-only, no migrations or funds paths. Rethrow keeps BullMQ retry behavior. Data shape change is only `components: {}` instead of a missing key on strategy `estimated`, which the schema and REST layer already expect. `projectStrategies` can include queue strategies without local `StrategyChanged` logs (matches on-chain queue; production with full history already had those events).

### DIFF
--- a/packages/ingest/abis/yearn/3/vault/snapshot/hook.spec.ts
+++ b/packages/ingest/abis/yearn/3/vault/snapshot/hook.spec.ts
@@ -61,4 +61,64 @@ describe('abis/yearn/3/vault/snapshot/hook', function() {
       globalThis.fetch = originalFetch
     }
   })
+
+  it('parses composition when estimated apr rows have only netAPR/netAPY components', async function() {
+    const originalFetch = globalThis.fetch
+    globalThis.fetch = (async () => ({
+      json: async () => []
+    })) as unknown as typeof fetch
+
+    const chainId = 1337
+    const vault = '0x3000000000000000000000000000000000000003'
+    const strategy = '0x4000000000000000000000000000000000000004' as `0x${string}`
+    const latest = Math.floor(Date.now() / 1000)
+
+    const outputs = [
+      { component: 'netAPR', value: 0.03 },
+      { component: 'netAPY', value: 0.031 }
+    ]
+
+    for (const output of outputs) {
+      const outputData = {
+        chain_id: chainId,
+        address: strategy,
+        label: 'katana-estimated-apr',
+        component: output.component,
+        value: output.value,
+        block_number: latest,
+        block_time: latest,
+        series_time: latest
+      }
+      await db.query(toUpsertSql('output', 'chain_id, address, label, component, series_time', outputData), Object.values(outputData))
+    }
+
+    const debts = [{
+      strategy,
+      activation: 0n,
+      lastReport: 0n,
+      currentDebt: 1n,
+      currentDebtUsd: 1,
+      maxDebt: 1n,
+      maxDebtUsd: 1,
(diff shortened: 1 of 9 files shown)


murderteeth (approved):
## Summary

Fixes stale Katana vault snapshots by ensuring estimated APR data always includes the schema-required `components` field. The accompanying regression coverage passes.

## Issues

None.

## Suggestions

- **`packages/ingest/yvusd-estimated-apr.containers.spec.ts:187`** — Stop the ingest container before deleting orphan vault things. Otherwise, historical replay could recreate an orphan between the `DELETE` and `refresh-vaults`, causing an intermittent Zod failure. Mirror the ordering already used in the Katana spec by adding `await result.ingestContainer?.stop()` immediately before the orphan-thing `DELETE`. This is non-blocking and can be handled in a follow-up after PR #443 merges.

## Verdict

APPROVE

---

## How This Was Reviewed

PR #445, opened 2026-07-26
Add yTranche indexing and REST support

### LINKED ISSUE #444: Add yTranche indexing and REST support
## Description

Add support for the current Ethereum yTranche deployment. Kong will discover tranches through the controller, retain controller state, enrich tranche snapshots with current Hook state, generate controller-backed tranche PPS and performance history, and expose the data through REST only.

## Scope

Deployment:

| Component | Address |
|---|---|
| Main vault | `0xDa87123895a043Ed3610155550177C54ce8ba49B` |
| Controller | `0xF0145433E5289dd10712650dCd28333FA317eF36` |
| Hook | `0x776DEd3273440f1481d07B6CE916b5d5Fac170dC` |

A controller is bound to one asset class: `ASSET` and `VAULT` are constructor arguments with no setters, so a USD, BTC and ETH system each require their own controller. Only the USD deployment above is indexed, and nothing may assume a chain has exactly one controller.

Out of scope:

- GraphQL changes
- Shareholder balances or cooldown state
- Indexing additional deployments or chains — though everything keyed by controller must admit more than one per chain
- Hook things or standalone Hook snapshots
- Hook-change tracking
- New event hooks
- New REST performance shapes
- Governance and transaction execution

## Internal naming

Use functional names without the `y` prefix:

- `trancheController`
- `tranche`

Use “yTranche” in documentation and user-facing descriptions.

### 1. Index the deployment

Add controller and tranche ABIs as sibling paths:

```text
packages/ingest/abis/yearn/3/tranche/controller/
packages/ingest/abis/yearn/3/tranche/vault/
packages/ingest/abis/yearn/3/tranche/hook/
```

They must be siblings rather than sharing a `yearn/3/tranche` hook path: kong's hook resolver applies a parent path's hooks to every child path, which would cross-wire the controller and the tranches.

Include the Hook read interface required by the tranche snapshot hook. Configure the Ethereum controller as a static source in `config/abis.yaml`.

Events contained in the configured ABIs must be indexed through Kong’s existing automatic event extraction. Do not add event hooks.

### 2. Discover tranches

Add a controller snapshot hook that:

- Reads `VAULT`, `ASSET`, `reserveVault`, `tranchesLength`, and `tranchesByPriority`.
- Discovers every registered tranche in priority order.
- Creates or updates the required `tranche`, `vault`, and `strategy` things.
- Creates or updates a `trancheController` thing carrying `asset`, `mainVault`, `inceptBlock` and `inceptTime`, whether or not any tranches are registered. Deployments are enumerated from this label rather than from the tranches pointing at it.
- Adds these defaults to each tranche:

```text
trancheController
priority
trancheType
asset
mainVault
inceptBlock
inceptTime
erc4626
v3
yearn
```

`trancheType` identifies whether the contract uses the base or locked tranche implementation. A/B/E names are not used as protocol types.

The indexing configuration must ensure each tranche address has one snapshot owner and is not independently processed by conflicting generic ABI paths.

### 3. Enrich the controller snapshot with tranche state

Kong’s automatic controller snapshot stores the controller’s zero-argument state.

The controller snapshot hook must use `tranchesLength` and `tranchesByPriority(index)` to append an ordered `tranches` array. For each tranche, read `tranches(tranche)`, `liveAssets(tranche)`, and `trancheCoverage(tranche)`.

Each tranche entry must contain:

```text
address
priority
registered
accrualPaused
excessShareBps
targetRatePerSecondWad
baselineAssets
lastAccrual
pendingExcess
liveAssets
claim
covered
```

Use the automatic snapshot’s block for every appended read.

### 4. Enrich tranche snapshots with Hook state

Kong’s automatic tranche snapshot stores the current Solidity `hook()` return value as the raw `hook` address.

The tranche snapshot hook must use that address to append a separate `hookState` object containing:

```text
open
rateLimitWindow
depositLimit
depositRateLimit.used
depositRateLimit.windowStart
depositRateLimit.rateLimit
withdrawRateLimit.used
withdrawRateLimit.windowStart
withdrawRateLimit.rateLimit
depositCap
withdrawCap
```

Requirements:

- Read parameterized Hook properties using the tranche address as the target.
- `depositLimit` is the Hook's `depositLimits(tranche)` value.
- Preserve `hook` as the raw Hook address.
- Store derived Hook properties under `hookState`; do not return a second top-level `hook` field from the snapshot hook.
- Use the same block as the tranche snapshot.
- Store fixed-window counters as reported.
- The REST response may additionally expose effective usage after applying the snapshot timestamp and `rateLimitWindow`.
- Do not create a Hook thing.
- Do not store a standalone Hook snapshot.
- Do not index account-specific `allowed` or cooldown mappings.

### 5. Add accounting time series

Generate one daily `tranche-accounting` series per tranche:

```text
baselineAssets
pendingExcess
liveAssets
claim
covered
coverageRatio
targetRatePerSecondWad
excessShareBps
accrualPaused
```

Generate one daily `tranche-system` series at the controller address:

```text
totalClaims
vaultAssets
reserveAssets
backingAssets
coverageRatio
```

Values must be read at a single historical block for each observation. `coverageRatio` is null when there is no claim to cover — no claim is not evidence of full coverage.

### 6. Support controller-backed assets, PPS, and TVL-C

Refactor the existing vault calculations to obtain the authoritative asset balance through a shared function that accepts a vault thing and block number.

For ordinary vaults, read:

```text
vault.totalAssets()
```

When the vault thing has a `trancheController` default, read:

```text
trancheController.liveAssets(vault)
```

Use the authoritative asset balance in the existing `tvl-c` calculation.

For ordinary vault PPS, continue reading:

```text
vault.pricePerShare()
```

For tranche PPS, calculate:

```text
authoritativeAssets × 10^vaultDecimals / vault.totalSupply()
```

When `totalSupply` is zero, return `10^vaultDecimals`.

Use the resulting PPS reader for every current, weekly, monthly, and inception observation required by:

```text
pps
apy-bwd-delta-pps
```

`trancheController.liveAssets(vault)` already excludes `pendingExcess`; do not add pending excess to PPS or TVL.

Keep the existing output labels, components, pricing, historical sampling, annualization, compounding, storage, and REST shapes unchanged. Existing non-tranche vault calculations must remain unchanged.

### 7. Add REST resources

Add a collection and a member endpoint:

```text
GET /api/rest/tranche/:chainId/controllers
GET /api/rest/tranche/:chainId/:controller
```

The collection returns every system on the chain, one document per controller. The member returns one, addressed by controller. `controllers` is a static route segment, so it resolves ahead of the member route.

Each document returns:

```text
controller
asset
mainVault
reserveVault
system accounting
ordered tranches
blockNumber
blockTime
```

Each tranche entry must include its thing metadata, current `hook` address, `hookState`, controller accounting, current PPS, coverage, and effective deposit and withdrawal capacity.

Expose the new historical labels through Kong’s existing timeseries REST mechanism:

```text
tranche-accounting
tranche-system
pps
apy-historical
```

Do not add GraphQL fields or resolvers.

### 8. Prevent double counting

Tranche TVL represents claims on the main vault’s backing, not additional protocol assets.

- Emit tranche `tvl-c` outputs using `trancheController.liveAssets(tranche)`.
- Do not emit the legacy `tvl` label for tranche vaults.
- Do not create or calculate an aggregate that sums main-vault TVL and tranche TVL as independent backing.
- Do not include `pendingExcess` in live tranche NAV or TVL.
- Keep accounting coverage separate from immediately deliverable withdrawal capacity.

### 9. Tests and documentation

Add focused tests using Kong’s existing unit and fixed-block RPC test patterns:

- Verify authoritative asset source selection: ordinary vaults use `totalAssets()` and vault things with `trancheController` use `liveAssets(vault)`.
- Verify controller-backed assets against the deployed controller at a fixed block.
- Verify tranche PPS against `authoritativeAssets × scale / totalSupply` at a fixed block.
- Verify zero-supply PPS returns the share scale when the calculation is implemented as a pure helper.
- Verify ordinary vault PPS continues using `pricePerShare()`.
- Verify APY obtains its current, weekly, monthly, and inception PPS observations through the shared PPS reader.
- Verify `tvl-c` consumes the authoritative asset balance while preserving its existing output shape.
- Verify tranche vaults do not emit the legacy `tvl` label.
- Verify controller discovery returns the deployed tranche addresses in priority order at a fixed block.
- Verify tranche snapshot enrichment preserves the raw `hook` address and appends representative limit data under `hookState`.

Do not add new testing infrastructure or broad container-level tests for this work.

Document the REST endpoint, timeseries labels, accounting semantics, snapshot naming distinction, and current deployment.

Reference: [yTranche contracts](https://github.com/Schlagonia/ytranche)

## Acceptance criteria

- [ ] Kong discovers all current tranches from the controller without hardcoded tranche addresses.
- [ ] The controller snapshot is retained with block metadata.
- [ ] The controller has a `trancheController` thing, and REST enumerates deployments from it.
- [ ] Each tranche has trancheController, priority, and type metadata.
- [ ] Each tranche snapshot retains the raw `hook` address and enriched `hookState`.
- [ ] No Hook thing or standalone Hook snapshot is created.
- [ ] Controller accounting and backing series are generated daily.
- [ ] Tranche PPS and historical APY use controller-backed PPS.
- [ ] Tranche `tvl-c` uses controller-backed assets and does not emit legacy `tvl`.
- [ ] Existing PPS, APY, and REST output shapes remain unchanged.
- [ ] yTranche data is available through REST without GraphQL changes.
- [ ] Events are indexed without event hooks.
- [ ] Shareholder cooldown state is not indexed.
- [ ] Tranche claims are not counted as additional protocol TVL.
- [ ] Existing non-tranche vault behavior remains unchanged.

murderteeth:
### Summary

Indexes the yTranche deployment: Kong discovers tranches through the controller,
retains controller state, enriches tranche snapshots with Hook state, generates
controller-backed PPS and performance history, and exposes it over REST.

The controller is the authoritative source for a tranche's assets and price per
share — it holds what the tranche has accrued and what losses it has absorbed,
where the tranche's own `totalAssets()` reports its ERC-4626 balance. So `tvl-c`,
`pps` and every APY observation resolve through one shared reader:
`trancheController.liveAssets(vault)` for tranches, `totalAssets()` and
`pricePerShare()` for ordinary vaults, which keep the paths they already had.

Closes #444

### How to review

Read the commits in order — each is one self-contained step, and the messages carry
the reasoning that doesn't fit in code comments.

Start with `docs/ytranche.md`: it covers the accounting semantics that are easy to
get wrong (pendingExcess is assigned profit awaiting a report, coverage is an
accounting measure while capacity is what the Hook will deliver, rate limits are
fixed windows that expire at `windowStart + rateLimitWindow`) and the `hook` vs
`hookState` snapshot distinction.

Then the three places where a wrong call would be expensive:

- `packages/ingest/abis/yearn/lib/assets.ts` — the shared asset/PPS readers, and
  the refactor of `lib/tvl.ts`, `lib/apy.ts` and the v2/v3 `pps` hook onto them.
  This is the only change that reaches non-tranche vaults.
- `.../tranche/controller/snapshot/hook.ts` — discovery and the ordered `tranches`
  accounting array, all read at the automatic snapshot's block.
- `packages/web/app/api/rest/tranche/db.ts` — payload units and derivations.

The two generated `abi.ts` files are etherscan-verified sources, one line each —
skim or skip.

### Test plan

- [x] Automated: `bun --filter ingest test`, `bun --filter web test`, `bun --filter lib test`.
      20 new tests. Ingest carries 10 failures that also fail on `main`
      (archive-RPC-dependent specs) — the same set before and after this branch.
- [x] Manual: full local run against mainnet, controller-only config, discovery
      through to REST. Three tranches found in priority order; 9 things with correct
      base/locked types; `hookState` on each tranche alongside the raw `hook`
      address; five output labels populated with the legacy `tvl` label appearing
      for the main vault alone; both REST routes 200 with matching payloads; unknown
      controller 404, bad chainId 400.
- [x] Verified two `tranche-accounting` / `tranche-system` observations against
      direct `eth_call` at their stored block numbers — exact to the wei, confirming
      each observation is a point-in-time read.

### Risk / impact

Runs on the existing schema, reusing the `thing`, `snapshot` and `output` tables.

The one change with shared blast radius routes every yearn vault's `tvl-c` and `pps`
through the new readers. Ordinary vaults issue the same `totalAssets()` and
`pricePerShare()` calls they did before — the branch is keyed on a
`trancheController` default that tranches alone carry — and the existing `tvl` and
`pps` specs pass unchanged. Rollback is a straight revert; every write lands in
existing tables and columns.

Also adds a `refresh-cache` workflow step for the tranche cache and widens the
timeseries refresh to controller-addressed series. Both are additive; the tranche
endpoints begin serving once their first refresh runs.


rossgalloway (commented):
LGTM! 

Logic and rationale for how the system works looks sound. Expected output and data about the tranches and overall system look comprehensive and clear. Well documented and as easy to understand as possible for 2 complicated systems coming together. Nice work!

Below is the automated codex review with a few minor comments that I will let you determine whether are worth picking up before merge.

---
 ## Automated Review

  ## Summary

  Adds yTranche controller discovery, tranche and Hook-state snapshots, controller-backed asset/PPS calculations, accounting time series, and REST endpoints. Two correctness
  issues remain around null accounting semantics and stale Hook state.

  ## Issues

  - [ ] packages/web/app/api/rest/timeseries/db.ts:75 — The new accounting series cannot preserve the required coverageRatio: null when there is no claim. Both daily
    aggregation queries convert an all-null bucket to zero with COALESCE(AVG(NULLIF(value, 0)), 0).
      - Change: For coverageRatio in tranche-accounting and tranche-system, preserve SQL NULL in both getFullTimeseries and getRecentTimeseries; keep the existing aggregation
        and zero behavior for every pre-existing label and component.

      - Keep: Controller/tranche address scoping, route segments, legacy timeseries responses, and aggregation for other components must remain unchanged.
      - Done when: Latest and historical REST responses return coverageRatio: null for zero-claim observations in both new labels, with focused tests covering both queries.

  - [ ] packages/ingest/abis/yearn/3/tranche/vault/snapshot/hook.ts:39 — Removing a Hook or switching to an incompatible Hook retains the previous hookState. The hook returns
    {}, while upsertSnapshot merges that over the existing state; REST then pairs the current raw Hook address—including the zero address—with stale limits and capacities.
      - Change: Return an explicit hookState: null when the raw Hook is absent, zero, or cannot be enriched, and ensure the snapshot loader preserves that null as a clearing
        value.

      - Keep: Preserve the raw automatic hook address, keep enrichment failures non-fatal, and do not create a Hook thing or standalone Hook snapshot.
      - Done when: A valid snapshot followed by a zero-address or incompatible-Hook snapshot persists hookState: null, and REST returns null Hook state and capacity rather than
        the previous values.

  ## Suggestions

  - Consider extracting the duplicated scoped-series traversal from refresh.ts and refresh-historical.ts into a shared helper, while retaining their respective readers, Redis
    keys, and payload shapes.

  Coverage this review did not reach:

  - Controller-hook side effects that create/update controller, tranche, vault, strategy, and ERC-20 things—including a controller with zero tranches.
  - ABI/config integration, automatic event extraction, sibling hook resolution, and generic snapshot-owner exclusions.
  - End-to-end REST database, Redis, collection/member route, multi-controller, and 200/400/404 behavior.
  - Executable coverage of scoped timeseries discovery and cache population.
  - Non-tranche regression coverage across the broader V2/V3 and yield-bearing asset/PPS paths.

  ## Verdict

  COMMENT

  ———

  ## How This Was Reviewed

  Reviewed with the review-pr-workflow skill (https://github.com/yearn/webops/blob/main/skills/review-pr-workflow/SKILL.md) —
  5 review lenses, each material finding independently verified by fresh native Codex session agents. 0 candidate findings were refuted and dropped.

  ## Verification and preview

  - Lint: 0 errors, 34 existing/current warnings. The repository wrapper required unsupported terminal behavior, so its underlying package lint commands were run directly.
  - Web tests: 9/9 passed.
  - Ingest selection: 20/20 passed.

  - Production build: passed.
  - No UI files changed, so Playwright visual verification was skipped.
  - No source changes were made.

packages/ingest/abis/yearn/3/tranche/controller/snapshot/hook.ts
```diff
@@ -0,0 +1,226 @@
...
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export default async function process(chainId: number, address: `0x${string}`, data: any) {
+  const snapshot = SnapshotSchema.parse(data)
+  const { blockNumber } = snapshot
+
+  const addresses = await extractTranchesByPriority(chainId, address, snapshot.tranchesLength, blockNumber)
+  const tranches = await extractTrancheAccounting(chainId, address, addresses, blockNumber)
+
+  await projectControllerThing(chainId, address, snapshot)
```
matheus1lva, line 51: if we can do a promise.all here would be small tiny ways to make it faster.

packages/ingest/abis/yearn/lib/assets.ts
```diff
@@ -0,0 +1,98 @@
+import { EvmAddressSchema, Thing } from 'lib/types'
```
matheus1lva, line 1: the file contains mostly "tranche" related things, im wondering if doesnt make sense rename to be under lib/traches/assets maybe?

packages/web/app/api/rest/tranche/db.ts
```diff
@@ -0,0 +1,371 @@
...
+  controller: string,
+): Promise<TrancheSystem | null> {
+  const controllerRows = await db.query(`
+    SELECT
+      snapshot,
+      hook,
+      block_number AS "blockNumber",
+      block_time AS "blockTime"
+    FROM snapshot
+    WHERE chain_id = $1 AND lower(address) = lower($2)
```
matheus1lva, line 166:
```suggestion
    WHERE chain_id = $1 AND address = $2
```

and make sure controller is passed throug getAdress() from viem. All the addressess have been normalized already

packages/web/app/api/rest/tranche/db.ts
```diff
@@ -0,0 +1,371 @@
...
+  // No controller snapshot yet means nothing to serve — the tranches' own
+  // accounting lives in that row.
+  if (controllerRows.rows.length === 0) return null
+
+  const { snapshot, hook, blockNumber, blockTime } = controllerRows.rows[0]
+  const assetAddress = snapshot?.ASSET ?? null
+
+  const assetRows = assetAddress
+    ? await db.query(
+      'SELECT defaults FROM thing WHERE chain_id = $1 AND lower(address) = lower($2) AND label = $3',
```
matheus1lva, line 178: prob same here

matheus1lva (commented):
## Review

### Verdict: COMMENT

No blockers. Four confirmed issues (two extend the prior automated review with new detail, two are new), three non-blocking suggestions.

### Issues

- [ ] `packages/ingest/abis/yearn/3/tranche/vault/snapshot/hook.ts:39` — When a tranche's Hook is absent, zeroed, or unreadable, the hook returns `{}`; `upsertSnapshot` shallow-merges (`packages/ingest/load/index.ts:104-108`), so the previous `hookState` is retained. REST (`tranche/db.ts:254`, `:280-281`, `:302-310`) then serves the stale limits stamped with the *current* snapshot's blockNumber/blockTime, and derives `effectiveUsed`/`remaining`/`capacity` from them. `setHook` makes this reachable. Spec task 4 requires hookState read at the snapshot's block.
  - **Change:** Return `{ hookState: null }` for the absent/zero/incompatible cases so the merge overwrites the stored value (not `undefined` — it can be dropped in queue serialization), and have `buildTrancheSystem` treat null as absent (`hookState: null`, `capacity: null`, which `db.spec.ts:166-183` already exercises for the no-snapshot case). Add a test: populated state followed by a zero/incompatible-Hook snapshot persists null.
  - **Keep:** Raw `hook` address preserved; enrichment failures non-fatal; no second top-level `hook` field.
  - **Done when:** After a snapshot where the Hook is zero/unreadable, the stored row's `hook.hookState` is null and REST returns `hookState: null` / `capacity: null` instead of an older block's values.

- [ ] `packages/web/app/api/rest/timeseries/db.ts:75` (and `:103`) — `COALESCE(AVG(NULLIF(value, 0)), 0)` collapses an all-NULL daily bucket to 0. The SQL predates this PR, but #445 newly routes `tranche-accounting`/`tranche-system` through it, so "no claim to cover" is served as "0% covered" — the exact inversion spec task 5 forbids, and this PR's own docs (docs/outputs.md, docs/ytranche.md, docs/rest.md's `?components=coverageRatio`) promise null. Ingest is correct (`value: claim > 0n ? div(covered, claim) : undefined` persists as NULL); the aggregation destroys it.
  - **Change:** Preserve NULL through aggregation for these series — `AVG(value)` in both `getFullTimeseries` and `getRecentTimeseries`; if the COALESCE must stay for existing vault labels, branch on the label.
  - **Keep:** Existing vault labels (`pps`, `apy-bwd-delta-pps`, `apr-oracle`, `tvl-c`) keep current bucket values and shape; daily bucketing and the historical/latest merge unchanged.
  - **Done when:** `GET /api/rest/timeseries/tranche-accounting/1/<tranche>?components=coverageRatio` returns `value: null` for a zero-claim day, and likewise for `tranche-system`.

- [ ] `packages/ingest/abis/yearn/3/tranche/controller/snapshot/hook.ts:132` — `tranches.map((address, priority) => …)` derives `priority` from the caller's array index, not from the controller. `tranche-accounting/hook.ts:46` calls `extractTrancheAccounting(chainId, controller, [address], blockNumber)` with a single address, so every tranche comes back `priority: 0`. The spec cements the wrong behavior: `hook.spec.ts:93-96` passes `[a, e]` and asserts `[0, 1]` while the discovery test above proves `e` sits at index 2. Currently latent (the accounting hook doesn't consume `priority`), but the field lies for any subset caller.
  - **Change:** Take `{ address, priority }[]` (the controller snapshot knows the real index from `extractTranchesByPriority`) and have `tranche-accounting/hook.ts:46` pass the tranche's stored `defaults.priority`. Update `hook.spec.ts:96` to expect real priorities.
  - **Keep:** The controller snapshot's `tranches` array stays priority-ordered with the same field names/values.
  - **Done when:** `extractTrancheAccounting` cannot produce a priority disagreeing with `tranchesByPriority(index)`; the `[a, e] -> [0, 1]` assertion is replaced by `[0, 2]`.

- [ ] `packages/ingest/abis/yearn/3/tranche/controller/snapshot/hook.ts:171` (and `:69`) — `estimateCreationBlock` — a bytecode binary search (~25 archive calls, only a 10s cache; `packages/lib/blocks.ts:98-127`) — runs unconditionally for the controller and every tranche on every snapshot cycle. The repo's convention guards it: `yearn/3/vault/snapshot/hook.ts:119` runs it only `if (! await things.exist(...))`. Three tranches ≈ ~100 archive `getBytecode` calls per controller snapshot for values that never change.
  - **Change:** Guard both calls with `things.exist(chainId, address, label)` and omit `inceptBlock`/`inceptTime` from `defaults` when the thing exists — `upsertThingDefaults` (`packages/ingest/db.ts:124`) is a right-wins merge, so omitting keys preserves stored values.
  - **Keep:** Tranche things still re-project `priority`, `trancheType`, and the rest of `defaults` each snapshot; only incept fields are skipped.
  - **Done when:** A repeat snapshot of an already-discovered controller issues no `estimateCreationBlock` calls and stored incept values are unchanged.

### Suggestions

- `.github/workflows/refresh-cache.yml:31` — The refresh-cache step and the widening of `refresh.ts`/`refresh-historical.ts` to tranche/controller-addressed series are not requested in #444 (task 7 scopes to the two REST resources and existing timeseries REST). They are necessary plumbing — the routes read only Redis, so without them every tranche route 404s — and the PR body discloses them. Worth an explicit scope acknowledgment since they touch infrastructure shared with the vault caches.
- `packages/web/app/api/rest/tranche/db.ts:302` — `capacity` re-emits three `hookState` fields verbatim under different names (`depositCap`, `depositLimit`, `withdrawCap`) and ignores `remaining`, the rate-limit-aware number computed just above in `toRateLimit` (db.ts:122). Either drop `capacity`, or make it carry only derived deliverable numbers (e.g. `Math.min(cap, remaining)`) and stop restating `limit`.
- `packages/web/app/api/rest/timeseries/refresh-historical.ts:52` — The 34-line scoped-refresh pass is duplicated byte-for-byte with `refresh.ts:52-82`, differing only in fetcher and key function. Extract one `refreshScopedSeries(fetch, key)` helper so adding a fourth scoped label edits one place, not two.

**Coverage this review did not reach** (non-blocking): controller-discovery side effects (`process`/`projectControllerThing`/`projectTrancheThings` are never called by any spec, so zero-tranche controllers and thing/default projection are untested); ABI/config wiring (abis.yaml, sibling-path isolation, snapshot ownership, automatic event extraction); the non-tranche regression surface of `lib/assets.ts`/`tvl.ts`/`apy.ts` and the v2 PPS hook (no legacy-fallback or failed-read pinning); tranche-path historical APY (apy-shared-pps.spec uses an ordinary v2 vault); the REST DB layer beyond the pure payload builder (`getTrancheControllers`/`getTrancheSystems`, multi-controller, case handling); the two routes and cache refresh (no 200/400/404, header, or Redis-failure tests); timeseries label routing/refresh plumbing regression; and docs-vs-code cross-check of the negative-scope criteria.

---

## How This Was Reviewed
5 review lenses (skeptic / architect / minimalist models per agent-and-subagents rules),
each material finding independently verified (auto panel: codex gpt-5.6-terra). 2 candidate findings were refuted and dropped.

PR #447, opened 2026-07-29
fix(ingest): surface legacy v2 strategy debts and harden strategy projection

### LINKED ISSUE #446: Legacy V2 snapshots suppress live strategy debts and can project stale strategy lists
## Impact

Legacy Yearn V2 vault snapshots can report nonzero `totalDebt` while returning no strategy allocation data. Consumers see either no allocation at all or a stale/incomplete top-level `strategies` list.

This is directly visible for two still-funded Ethereum V2 vaults:

- USDC 0.3.0: the active router is first in `withdrawalQueue` and carries essentially all debt, but `strategies` contains an older residual-debt strategy and both `composition` and `debts` are empty.
- DAI 0.3.0: the active router is present in `strategies`, but `composition` and `debts` are still empty.

Downstream effects include missing/incorrect strategy allocations in consumers such as Powerglove and inability for TVL consumers to reconcile nested vault debt. In particular, both routers deposit into 0.4.3 V2 vaults; counting the parent vault's assets and the destination vault independently without the debt relationship can double-count nested TVL.

## Reproduction

- USDC: https://kong.yearn.fi/api/rest/snapshot/1/0x5f18C75AbDAe578b483E5F43f12a39cF75b973a9
- DAI: https://kong.yearn.fi/api/rest/snapshot/1/0x19D3364A399d251E894aC732651be8B0E4e85001

Current user-visible fields:

| Vault | API version | `totalDebt` | `strategies` | `composition` / `debts` |
|---|---:|---:|---|---|
| USDC | 0.3.0 | `1715254032028` | `0xC0176...B64b` | `[]` / `[]` |
| DAI | 0.3.0 | `888427034461733866977765` | `0x3D653...f1Be` | `[]` / `[]` |

## Pinned on-chain verification

All reads below were pinned to Ethereum block **25,634,133**, timestamp **2026-07-28 21:53:35 UTC**, hash `0x595e531ad690908332799198dad57963ab1a22602d233cabc5886e01112a6670`.

The 0.3.0 `strategies(address)` getter was decoded using that release's 8-field `StrategyParams` layout (`performanceFee`, `activation`, `debtRatio`, `rateLimit`, `lastReport`, `totalDebt`, `totalGain`, `totalLoss`), not the 9-field layout introduced in 0.3.2.

### USDC 0.3.0

Vault: `0x5f18C75AbDAe578b483E5F43f12a39cF75b973a9`

`withdrawalQueue` (nonzero entries, in order):

1. `0x36822d0b11F4594380181cE6e76bd4986d46c389`
2. `0x7900c70a377f89DF29D1D1939469ae3B74c5B740`
3. `0x429F5D3549096ec25D6F5E9BbA193bdDE9131584`
4. `0x387fCa8d7e2e09655b4F49548607B55C0580fC63`
5. `0x2B1a6CB0168aA540ee8D853aB1d10d7A89d6351b`
6. `0x80af28cb1e44C44662F144475d7667C9C0aaB3C3`
7. `0x520BF095FA58cB3f68C18D01746041733A1f7b85`
8. `0x86Aa49bf28d03B1A4aBEb83872cFC13c89eB4beD`

Router `0x36822d0b11F4594380181cE6e76bd4986d46c389`:

- `name()`: `Router-yvUSDC-030-043`
- `vault()`: `0x5f18C75AbDAe578b483E5F43f12a39cF75b973a9`
- `yVault()`: `0xa354F35829Ae975e850e23e9615b11Da1B3dC4DE`
- destination `apiVersion()`: `0.4.3`
- vault `strategies(router).debtRatio`: `10000` (100%)
- vault `strategies(router).totalDebt`: `1715254022028` (1,715,254.022028 USDC)
- vault `totalDebt()`: `1715254032028` (1,715,254.032028 USDC)
- difference: `10000` raw units (0.01 USDC)

The projected-only older strategy `0xC0176FAa0e20dFf3CB6B810aEaE64ef271B1b64b` is not in the queue, has debt ratio `0`, and has residual `totalDebt = 10000`. Its debt plus the router debt exactly equals vault `totalDebt` at the verification block.

The active router was added at block 13,594,531 by the 0.3.0 four-argument `StrategyAdded(address,uint256,uint256,uint256)` event (tx `0xf91758609beaea754446a538b962e1fb37bd33b3302c941a7e0f7f28d71cbca9`). The older projected address entered via `StrategyMigrated` at block 13,397,022.

### DAI 0.3.0

Vault: `0x19D3364A399d251E894aC732651be8B0E4e85001`

`withdrawalQueue` (only nonzero entry):

1. `0x3D6532c589A11117a4494d9725bb8518C731f1Be`

Router `0x3D6532c589A11117a4494d9725bb8518C731f1Be`:

- `name()`: `Routeryvdai043`
- `vault()`: `0x19D3364A399d251E894aC732651be8B0E4e85001`
- `yVault()`: `0xdA816459F1AB5631232FE5e97a05BBBb94970c95`
- destination `apiVersion()`: `0.4.3`
- vault `strategies(router).debtRatio`: `9924` (99.24%)
- vault `strategies(router).totalDebt`: `888427034461733437210185` (888,427.034461733437210185 DAI)
- vault `totalDebt()`: `888427034461733866977765` (888,427.034461733866977765 DAI)
- difference: `429767580` wei (0.000000000429767580 DAI)

This router entered via `StrategyMigrated` at block 12,820,915 (tx `0x9b00053c09c105abd869cf3489b435906613aacb4bf3ae777b51b0ad2ceb6495`), which explains why Kong's event projector sees it even though it does not see the USDC router's old-signature `StrategyAdded` event.

## Relevant Kong path and confirmed causes

1. [`packages/ingest/abis/yearn/2/vault/snapshot/hook.ts` lines 72-77](https://github.com/yearn/kong/blob/0eab01307593cf1605c85111e01d07042fbac31a/packages/ingest/abis/yearn/2/vault/snapshot/hook.ts#L72-L77) explicitly sets `oldold = apiVersion <= 0.3.1` and unconditionally returns empty `debts` and `composition` for those releases. This confirms why both 0.3.0 examples are empty and shows the problem is not limited to 0.3.0: 0.2.2 and 0.3.1 are gated too.
2. [`projectStrategies()` lines 131-163](https://github.com/yearn/kong/blob/0eab01307593cf1605c85111e01d07042fbac31a/packages/ingest/abis/yearn/2/vault/snapshot/hook.ts#L131-L163) projects the top-level list from ingested `StrategyAdded`, `StrategyMigrated`, and `StrategyRevoked` logs. It subscribes only to the post-0.3.1 five-argument `StrategyAdded` signature. [V2 0.3.0 used a four-argument event and an 8-field strategy struct](https://github.com/yearn/yearn-vaults/blob/v0.3.0/contracts/Vault.vy#L91-L115); [0.3.2 introduced min/max debt-per-harvest and the five-argument event](https://github.com/yearn/yearn-vaults/blob/v0.3.2/contracts/Vault.vy#L91-L119). This confirms why the USDC router's direct add is invisible while the DAI router's migration is visible.
3. [`extractDebts()` lines 182-203](https://github.com/yearn/kong/blob/0eab01307593cf1605c85111e01d07042fbac31a/packages/ingest/abis/yearn/2/vault/snapshot/hook.ts#L182-L203) reads `hook->'strategies'` from the existing stored snapshot and then queries only those addresses. It does not receive the newly projected list or the current queue. This makes an incomplete/stale projected list directly suppress debt reads (and may add a refresh-cycle lag).
4. [`extractWithdrawalQueue()` lines 245-273](https://github.com/yearn/kong/blob/0eab01307593cf1605c85111e01d07042fbac31a/packages/ingest/abis/yearn/2/vault/snapshot/hook.ts#L245-L273) already reads all 20 queue slots successfully, but that authoritative current-state list is not used to build debts.
5. [`extractComposition()` lines 308-368](https://github.com/yearn/kong/blob/0eab01307593cf1605c85111e01d07042fbac31a/packages/ingest/abis/yearn/2/vault/snapshot/hook.ts#L308-L368) can distinguish active, inactive, and unallocated entries, but receives empty debts for old releases and iterates only the event-projected `strategies` list.

The REST route is not independently dropping these fields: the combined refresh serializes the assembled DB snapshot into cache ([`refresh-vaults.ts` lines 47-65](https://github.com/yearn/kong/blob/0eab01307593cf1605c85111e01d07042fbac31a/packages/web/app/api/rest/refresh-vaults.ts#L47-L65)).

## Scope scan (live REST, same investigation)

I enumerated all 638 entries whose `apiVersion` begins with `0.` from `GET /api/rest/list/vaults?origin=yearn`, fetched every corresponding snapshot (0 fetch errors), and selected snapshots with `totalDebt > 0`, `debts = []`, and `composition = []`.

**13 Ethereum V2 snapshots currently match:**

| API version | Count | Vaults |
|---|---:|---|
| 0.2.2 | 1 | `0xBFa4...ECBB` |
| 0.3.0 | 7 | `0xdCD9...4325`, `0x5f18...73a9`, `0x19D3...5001`, `0x0e8A...5f14`, `0xED02...55f6`, `0x0e88...5A47`, `0xB98D...BCEC` |
| 0.3.1 | 2 | `0xac33...d82B`, `0x3466...8DfC` |
| 0.3.3 | 1 | `0x477f...8126` |
| 0.3.5 | 1 | `0xF962...94c3` |
| 0.4.6 | 1 | `0x6A56...08aD` |

The first **10** are explained deterministically by the `<= 0.3.1` gate (10 of the 18 listed vaults in those releases currently have nonzero `totalDebt`). The three later-version matches are not version-gated; all have `strategies = []` but nonempty `withdrawalQueue`, so they appear to be additional event-projection/ingestion-gap candidates and should be checked while fixing the common discovery path.

## Likely correction (proposal / design hypothesis)

Use current on-chain state as the source of truth for current allocation, while keeping event projection for historical identity/status:

1. Select the correct `StrategyParams` ABI by V2 release:
   - 0.2.2 has `debtLimit` + `rateLimit`;
   - 0.3.0-0.3.1 have `debtRatio` + `rateLimit` (8 fields);
   - 0.3.2+ have `debtRatio` + min/max debt per harvest (9 fields).
2. Build candidates from the union of current nonzero `withdrawalQueue` entries and version-correct historical `StrategyAdded`/`StrategyMigrated` events. Do not let an incomplete event projection exclude a current queue entry.
3. Read `strategies(address)` for those candidates directly in the same snapshot pass, and pass the resulting set into debt/composition construction instead of re-reading the previous snapshot's `hook.strategies`.
4. Preserve distinctions:
   - currently allocated/debt-bearing strategies;
   - queued but zero-allocation strategies;
   - revoked or migrated historical strategies with residual debt;
   - zero-debt historical/unallocated strategies.
5. Deduplicate addresses before summing. Router debt is parent-vault debt; shares/assets inside the destination V2/V3 vault must be represented as a relationship for nested-TVL reconciliation, not added again to parent TVL.

The exact public schema/status naming is a design choice; the confirmed defect is that current debt-bearing allocations are absent.

## Acceptance criteria

- [ ] V2 0.2.2, 0.3.0, and 0.3.1 use version-correct event and `strategies(address)` layouts; remove the unconditional empty-allocation behavior.
- [ ] The USDC snapshot includes router `0x3682...c389` as current/active with 100% debt ratio and its current debt, while retaining `0xC017...B64b` as a non-current residual/historical strategy without double-counting.
- [ ] The DAI snapshot includes router `0x3D65...f1Be` with 99.24% debt ratio and current debt.
- [ ] Current nonzero queue entries cannot be omitted solely because their add event was not projected.
- [ ] Revoked/migrated/zero-allocation strategies remain distinguishable from active allocations; residual debt is not silently dropped.
- [ ] Sum of unique strategy debts reconciles to vault `totalDebt` within a documented tolerance; nested router destinations do not cause parent/child TVL double-counting.
- [ ] Add fixtures/tests for old 4-argument `StrategyAdded`, old 8-field `StrategyParams`, migration-only discovery, stale/missing projected strategies, and `splice(indexOf(...))` when the address is absent.
- [ ] Re-scan the 13 live matches above; resolve the 10 version-gated cases and classify/fix the three later projection-gap cases.

matheus1lva:
### Summary
Fixes #446. Legacy v2 vaults (apiVersion <= 0.3.1) snapshot with empty `debts`/`composition` and stale strategy lists. Three causes: the vault abi and event hook only knew the modern 5-arg `StrategyAdded` event (legacy vaults emit a 4-arg variant), the snapshot hook gated allocation extraction off entirely for <= 0.3.1, and debt candidates came from the previous snapshot's `hook.strategies` instead of live state. This PR ingests the 4-arg event, removes the gate, selects the 8- vs 9-field `strategies(address)` return layout by apiVersion, and builds debt candidates from projected events + live `withdrawalQueue` + revoked strategies. Revoked strategies stay listed only while they carry residual debt, so debt sums still reconcile to vault `totalDebt`.

Also fixes a latent bug in `projectStrategies`: revoking an address that was never projected made `splice(indexOf(...))` (`indexOf` = -1) drop the last element of the list. Address matching is now case-insensitive throughout.

### How to review
Start with `packages/ingest/abis/yearn/2/vault/snapshot/hook.ts`:
- `projectStrategyEvents` — event replay now tracks revoked separately and handles both `StrategyAdded` signatures
- `mapStrategyParams` — pre-0.3.2 vaults return 8 fields; pre-0.3.0 field 3 is `debtLimit` (absolute token amount), so it is not published as `debtRatio`
- `resolveStrategies` — which revoked strategies survive into the final list
- `extractComposition` — pre-0.3.0 vaults have no `debtRatio`, so queue membership stands in for the active check; >= 0.3.0 status semantics unchanged

The `abi.ts` change is one added event definition inside the flattened abi string; `event/StrategyAdded/hook.ts` adds the 4-arg topic.

Smoke test: snapshot a legacy vault (e.g. any 0.3.0/0.3.1 vault on mainnet) and confirm `debts` and `composition` are non-empty; snapshot a modern >= 0.3.2 vault and confirm output matches main.

### Test plan
- [ ] Manual: stride reset + refanout on an affected legacy vault, then verify snapshot `debts`/`composition` populate
- [ ] Automated: `bun --filter ingest test` — spec added covering `union`, `projectStrategyEvents` (both event signatures, revoke of unknown address), `resolveStrategies` residual-debt rules, and `mapStrategyParams` per-version field layouts

### Risk / impact
- Existing `evmlog_strides` cover the historical ranges, so affected legacy vaults need a stride reset + refanout before the 4-arg `StrategyAdded` events backfill — until then legacy vaults stay as they are today.
- Modern vaults (>= 0.3.2) now derive debt candidates from event projection + live queue instead of the prior snapshot's `hook.strategies`; output should be identical, but this is the path to watch for regressions.
- No migrations, no funds movement, no auth changes. Rollback is a revert; snapshots self-heal on the next run.


murderteeth (approved):
## Summary
Fixes #446. Legacy V2 vaults (apiVersion ≤ 0.3.1) snapshotted with empty `debts`/`composition` and stale `strategies`. The PR (1) adds the 4-arg `StrategyAdded(address,uint256,uint256,uint256)` event to the vault abi and the event hook's topics, (2) removes the `oldold` gate that unconditionally emptied debts/composition, (3) selects the 8- vs 9-field `strategies(address)` return layout by apiVersion, (4) builds debt candidates from event projection ∪ live `withdrawalQueue` ∪ revoked strategies instead of the previous snapshot's `hook.strategies`, and (5) fixes the `splice(indexOf(...))` bug that silently dropped the last list element when revoking an unknown address.

## Verification performed
- **Lint**: `bun --filter ingest lint` — 0 errors (32 pre-existing warnings, none in changed files).
- **Tests**: the changed spec passes (17/17). The full ingest suite has 10 failures in `apy.spec.ts`, `yearn/3/strategy/event/hook.spec.ts`, and `erc4626/timeseries/pps/hook.spec.ts` — re-running those specs on `origin/main` produces the identical 10 failures (RPC-environment related), so they are pre-existing and unrelated.
- **On-chain smoke test** (mainnet, pinned to the issue's verification block 25,634,133): decoding USDC 0.3.0 vault `strategies(router)` with the PR's 8-field layout returns `debtRatio=10000`, `totalDebt=1715254022028`; the residual strategy returns `totalDebt=10000`; the sum equals vault `totalDebt=1715254032028` exactly — matching #446's pinned values. The legacy event selector `0x5ec27a4f...` matches topic0 of the actual `StrategyAdded` log in tx `0xf917586...` that main's code misses.
- **Pipeline trace**: `getLogs` receives all abi events, so the added 4-arg event is fetched and stored with `signature = topics[0]`, which is exactly what `projectStrategyEvents` queries. `StrategyAdded` is not in `config/blacklist.yaml`. The snapshot extractor passes all no-arg view fields (`token`, `decimals`, `apiVersion`) in `data`, so `extractDebts`'s switch from the stored snapshot row to `data` is sound and removes a refresh-cycle lag.
- **Migration semantics** (prod evmlog, all chains): all 325 `StrategyMigrated` events have a same-block `StrategyRevoked(oldVersion)` — `migrateStrategy` revokes internally — so the PR's revoked/residual-debt rule correctly covers migrated-away strategies too, with no special-casing needed.
- **Acceptance criteria from #446**: all code-level criteria are met, including the queue-union rule ("queue entries cannot be omitted because their add event wasn't projected") — this also structurally fixes the three non-version-gated cases (0.3.3/0.3.5/0.4.6 with empty projections but populated queues), with no stride reset needed for those since the queue read is live. The re-scan of the 13 live matches remains a post-deploy manual step, correctly listed in the test plan.

## Issues
None blocking.

## Related issues found during review (pre-existing, not caused by this PR)
- #449 — `extractFees__v2` inflates strategist performance fees ~10,000× (missing debt-ratio normalization). Latent today; this PR's legacy backfill will project nonzero-fee legacy strategies into that path, so it's worth fixing before or alongside the stride reset.
- #450 — the legacy 8-arg `StrategyReported` is never ingested, leaving pre-0.3.2 vault APY null. Same gap class as this PR's `StrategyAdded` fix; a natural follow-up using the same pattern, ideally sharing the same stride reset.

## Verdict
APPROVE

---

## How This Was Reviewed

Issue #409, opened 2026-05-15
Strategy performance.estimated can show yvUSD-scoped data in non-yvUSD vaults


rossgalloway:
## Summary

The same strategy can show different `performance.estimated.apy` values in different vault snapshots because Kong hydrates strategy-level `yvusd-estimated-apr` rows into each vault's strategy composition independently, then caches/embeds that value in the vault snapshot.

This creates two problems:

1. A yvUSD-contextual strategy estimate can appear inside non-yvUSD vaults that use the same strategy address.
2. The same strategy can show different estimated APYs across vaults depending on when each vault snapshot was hydrated/refreshed.

## Example

Strategy:

```txt
Morpho Yearn OG USDC Compounder
0x0e297dE4005883C757c9F09fdF7cF1363C20e626
```

In yvUSD snapshot:

```txt
Vault: 0x696d02Db93291651ED510704c9b286841d506987
Endpoint: /api/rest/snapshot/1/0x696d02Db93291651ED510704c9b286841d506987

performance.estimated.apr: 0.0484355921070911
performance.estimated.apy: 0.0496243924374524
performance.estimated.type: yvusd-estimated-apr
components.debtRatio: 0.215459283724573
```

In USDC-2 snapshot:

```txt
Vault: 0xAe7d8Db82480E6d8e3873ecbF22cf17b3D8A7308
Endpoint: /api/rest/snapshot/1/0xAe7d8Db82480E6d8e3873ecbF22cf17b3D8A7308

performance.estimated.apr: 0.0596473238823395
performance.estimated.apy: 0.0614569555075071
performance.estimated.type: yvusd-estimated-apr
components.debtRatio: 0.215459283724573
```

These are the same strategy address, but the embedded estimated APY differs:

```txt
yvUSD:  4.9624%
USDC-2: 6.1457%
```

## Expected behavior

For a given strategy, the current strategy-level estimated APR/APY should either:

- be the same everywhere it is shown, if it is truly strategy-scoped; or
- only be shown in the vault context it was computed for, if it is vault/context-scoped.

A `yvusd-estimated-apr` value, especially one with a `debtRatio` component from yvUSD, should not appear as if it is a generic estimate for the same strategy inside unrelated vaults.

## Actual behavior

Kong appears to hydrate strategy composition rows by matching output rows on strategy address.

The yvUSD APR service emits rows keyed by:

- vault address for yvUSD vault-level APR/APY
- strategy address for yvUSD strategy-level APR/APY/debt ratio

[...]

PR #451, opened 2026-07-31
fix(web): serve contract state over stale hook keys

matheus1lva:
### Summary

REST vault list and snapshot responses could serve a stale top-level pricePerShare from snapshot.hook. Vaults first indexed through the erc4626 ABI stored the PPS sparkline value in that blob. When yearn/3/vault took over, the old key remained, and the previous merge order (defaults < snapshot < hook) let it override the current multicall value.

This PR adds mergeSnapshot, which merges defaults < hook < snapshot and then restores the hook-enriched asset object when the hook holds a non-null asset; a null or missing hook asset falls back to contract state. It is used in the REST vault list/snapshot cache builder (getVaultsWithSnapshots) and in the nine GraphQL resolvers that build entities from the same three blobs (vault, vaults, accountVaults, strategy, strategies, vaultStrategies, accountStrategies, accountant, accountants).

One data-at-rest fix is included: the yearn/3/vault snapshot hook now echoes pricePerShare from contract state, so the stale hook key gets overwritten naturally on the next snapshot pass — no migration needed. erc4626-only vaults, whose state has no pricePerShare, keep serving their hook value.

Also removes an unused map index param in LatestBlocks.tsx, restyles the erc4626 PPS spec to Vitest timeout options, and gates four ingest specs that hit polygon/base RPC with it.skipIf on the configured chains, so mainnet-only chains.local.yaml runs skip them instead of failing while CI's full chains.yaml still runs them.

### How to review

- packages/web/lib/mergeSnapshot.ts: merge order and the asset exception.
- packages/web/app/api/rest/list/db.ts and packages/web/app/api/gql/resolvers/*: the call sites. The strategies resolver keeps the pre-existing falsy short-circuit on the apiVersion filter, so empty-string apiVersion rows are still kept.
- packages/web/app/api/rest/snapshot/db.ts: VaultSnapshot is now an alias of SnapshotRow.
- packages/ingest/abis/yearn/3/vault/snapshot/hook.ts: the pricePerShare echo.
- packages/web/lib/mergeSnapshot.spec.ts: unit coverage for precedence, asset preservation, and missing blobs.
- packages/ingest/yvusd-estimated-apr.containers.spec.ts: the seeded stale-hook suite (Postgres, no RPC) covers REST snapshot, REST list, the GraphQL vault query, and the cached Redis payloads (snapshot + list keys), next to the issue #409 suite.
- Additional ingest specs cover v3 PPS hook delegation, the snapshot hook echo, snapshot multicall field extraction, and ABI field discovery.

### Automated tests

- [x] bun --filter web test: 16 tests passed (incl. lib/mergeSnapshot.spec.ts).
- [x] packages/ingest unit suite: 22 files passed, 109 tests passed, 11 skipped (chain and env gates).
- [x] bun --filter ingest test:containers -- yvusd-estimated-apr.containers.spec.ts: 8 tests passed (3 issue-#409 + 5 stale-hook). One earlier run flaked in the issue-#409 beforeAll — in-flight ingest re-created orphan name-less vault things between the cleanup DELETE and refresh-vaults; pre-existing race in that suite's setup, rerun green.

### Manual test

Reproduces the stale-hook suite in yvusd-estimated-apr.containers.spec.ts by hand, against yvUSD on mainnet.

1. Config overrides — create these three files:

   `config/chains.local.yaml`
   ```yaml
   chains: ['mainnet']
   ```

   `config/abis.local.yaml`
   ```yaml
   abis:
     - abiPath: yearn/3/vault
       sources:
         - chainId: 1
           address: '0x696d02Db93291651ED510704c9b286841d506987'
           inceptBlock: 24271831
   ```

   `config/manuals.local.yaml`
   ```yaml
   manuals:
     - chainId: 1
       address: '0x696d02Db93291651ED510704c9b286841d506987'
       label: vault
       defaults: { inceptBlock: 24271831, origin: yearn, apiVersion: '3.0.4' }
   ```

2. Run ingest: `make dev`, then in the terminal UI select `ingest` → `fanout abis`. Re-run fanout until the snapshot row lands:
   ```sql
   SELECT snapshot->>'pricePerShare' FROM snapshot
   WHERE chain_id = 1 AND address = '0x696d02Db93291651ED510704c9b286841d506987';
   ```
   Should show the fresh multicalled value (uint256 string, e.g. `1021955...`).

3. Plant the erc4626-era stale key into the hook blob (the prod row shape the bug depends on):
   ```sql
   UPDATE snapshot SET hook = hook || '{"pricePerShare": 1002336}'::jsonb
   WHERE chain_id = 1 AND address = '0x696d02Db93291651ED510704c9b286841d506987';
   ```

4. Rebuild the REST cache: `bun packages/web/app/api/rest/refresh-vaults.ts` (web must be running — `make dev` starts it on :3001).

5. REST snapshot:
   ```
   curl -s localhost:3001/api/rest/snapshot/1/0x696d02db93291651ed510704c9b286841d506987 | jq '{pricePerShare, asset}'
   ```
   pricePerShare must be the fresh state value from step 2, NOT `1002336`; asset must stay the hook-enriched erc20 object (symbol/decimals), not a raw address.

6. REST list:
   ```
   curl -s localhost:3001/api/rest/list/vaults | jq '.[] | select(.address | ascii_downcase == "0x696d02db93291651ed510704c9b286841d506987") | .pricePerShare'
   ```
   Same fresh value.

7. GraphQL (http://localhost:3001/api/gql):
   ```graphql
   { vault(chainId: 1, address: "0x696d02Db93291651ED510704c9b286841d506987") { pricePerShare asset { symbol decimals } } }
   ```
   Same fresh value, enriched asset.

8. On main, steps 5–7 return `1002336` instead — that's the bug.

### Risk / impact

The merge change covers the REST caches and GraphQL. When snapshot and hook keys overlap, snapshot state wins except for asset, which stays hook-enriched. For vaults that previously served the stale hook value, pricePerShare changes from a number to the uint256 string form contract state uses; consumers should already tolerate both since pure-erc4626 vaults keep serving a number. Stale hook keys are not deleted at rest; the v3 snapshot hook echo overwrites them on the next snapshot pass, and reverting the code simply restores the previous merge behavior.

### DIFF
--- a/packages/ingest/abis/erc4626/timeseries/pps/hook.spec.ts
+++ b/packages/ingest/abis/erc4626/timeseries/pps/hook.spec.ts
@@ -2,11 +2,12 @@ import { expect } from 'chai'
 import { base, mainnet } from 'viem/chains'
 import { _compute } from './hook'
 import { ThingSchema } from 'lib/types'
+import { chains } from 'lib'
 
-describe('abis/erc4626/timeseries/pps/hook', function() {
-  it('extracts sdai pps', async function() {
-    this.timeout(30_000)
+const hasBase = chains.some(chain => chain.id === base.id)
 
+describe('abis/erc4626/timeseries/pps/hook', () => {
+  it('extracts sdai pps', { timeout: 30_000 }, async () => {
     const sdai = '0x83F20F44975D03b1b09e64809B757c47f942BEeA'
     const vault = ThingSchema.parse({
       chainId: mainnet.id,
@@ -18,9 +19,7 @@ describe('abis/erc4626/timeseries/pps/hook', function() {
     expect(pps.humanized).to.be.closeTo(1.1040043785400659, 1e-5)
   })
 
-  it('extracts avantis usdc pps', async function() {
-    this.timeout(30_000)
-
+  it.skipIf(!hasBase)('extracts avantis usdc pps', { timeout: 30_000 }, async () => {
     const usdc = '0x944766f715b51967E56aFdE5f0Aa76cEaCc9E7f9'
     const vault = ThingSchema.parse({
       chainId: base.id,
@@ -33,9 +32,7 @@ describe('abis/erc4626/timeseries/pps/hook', function() {
     expect(pps.humanized).to.be.closeTo(1.2807073904123205, 1e-5)
   })
 
-  it('uses share decimals when they differ from asset decimals', async function() {
-    this.timeout(30_000)
-
+  it('uses share decimals when they differ from asset decimals', { timeout: 30_000 }, async () => {
     // Yearn USDT (Morpho): 18-decimal shares, 6-decimal USDT asset.
     // defaults.decimals holds the asset decimals (6); convertToAssets needs the
     // share decimals (18) or it rounds to 0. Pre-fix this returned 0.
--- a/packages/ingest/abis/yearn/3/strategy/event/hook.spec.ts
+++ b/packages/ingest/abis/yearn/3/strategy/event/hook.spec.ts
@@ -1,6 +1,10 @@
 import { expect } from 'chai'
+import { polygon } from 'viem/chains'
+import { chains } from 'lib'
 import { HarvestSchema, computeApr, totalAssets } from './hook'
 import { addresses } from '../../../../../test-addresses'
+
+const hasPolygon = chains.some(chain => chain.id === polygon.id)
 
 function mock() {
   return HarvestSchema.parse({
@@ -18,7 +22,7 @@ function mock() {
 }
 
 describe('abis/yearn/3/strategy/event/hook', function() {
-  it('extracts totalDebt', async function() {
+  it.skipIf(!hasPolygon)('extracts totalDebt', async function() {
     const debt = await totalAssets(mock())
     expect(debt).to.equal(10489089449n)
   })
@@ -35,7 +39,7 @@ describe('abis/yearn/3/strategy/event/hook', function() {
     expect(apr.net).to.equal(0)
   })
 
-  it('computes gross and net apr on profit', async function() {
+  it.skipIf(!hasPolygon)('computes gross and net apr on profit', async function() {
     const latest = {
       ...mock(),
       blockNumber: 51916666n,
@@ -53,7 +57,7 @@ describe('abis/yearn/3/strategy/event/hook', function() {
     expect(apr.net).to.equal(0.04795259723804295)
   })
 
-  it('computes gross and net apr on loss', async function() {
+  it.skipIf(!hasPolygon)('computes gross and net apr on loss', async function() {
     const latest = {
       ...mock(),
       blockNumber: 51916666n,
--- a/packages/ingest/abis/yearn/3/vault/snapshot/hook.spec.ts
+++ b/packages/ingest/abis/yearn/3/vault/snapshot/hook.spec.ts
@@ -1,5 +1,5 @@
 import { expect } from 'chai'
-import { extractComposition } from './hook'
+import process, { extractComposition } from './hook'
 import db, { toUpsertSql } from '../../../../../db'
 
 describe('abis/yearn/3/vault/snapshot/hook', function() {
@@ -117,6 +117,35 @@ describe('abis/yearn/3/vault/snapshot/hook', function() {
       expect(composition[0].performance?.estimated?.apr).to.equal(0.03)
       expect(composition[0].performance?.estimated?.apy).to.equal(0.031)
       expect(composition[0].performance?.estimated?.components).to.deep.equal({})
+    } finally {
+      globalThis.fetch = originalFetch
+    }
+  })
+
+  it('echoes snapshot pricePerShare to override stale hook keys', async function() {
+    const originalFetch = globalThis.fetch
+    globalThis.fetch = (async () => ({
+      json: async () => []
+    })) as unknown as typeof fetch
+
+    const chainId = 1337
+    const vault = '0x5000000000000000000000000000000000000005'
+    const asset = '0x6000000000000000000000000000000000000006'
+    const pricePerShare = 1021955n
+    const assetData = {
+      chain_id: chainId,
+      address: asset,
+      label: 'erc20',
+      defaults: { name: 'USD Coin', symbol: 'USDC', decimals: 6 }
+    }
+    await db.query(toUpsertSql('thing', 'chain_id, address, label', assetData), Object.values(assetData))
+
+    try {
+      const composition = await extractComposition(chainId, vault, [], [])
+      const hook = await process(chainId, vault, { asset, pricePerShare })
+
+      expect(composition).to.have.length(0)
+      expect(hook.pricePerShare).to.equal(pricePerShare)
     } finally {
       globalThis.fetch = originalFetch
     }
--- a/packages/ingest/abis/yearn/3/vault/snapshot/hook.ts
+++ b/packages/ingest/abis/yearn/3/vault/snapshot/hook.ts
@@ -88,7 +88,8 @@ export const SnapshotSchema = z.object({
   accountant: EvmAddressSchema.optional(),
   role_manager: EvmAddressSchema.optional(),
   use_default_queue: z.boolean().optional(),
-  get_default_queue: EvmAddressSchema.array().optional()
+  get_default_queue: EvmAddressSchema.array().optional(),
+  pricePerShare: z.bigint({ coerce: true }).optional()
 })
 
 type Snapshot = z.infer<typeof SnapshotSchema>
@@ -159,6 +160,9 @@ export default async function process(chainId: number, address: `0x${string}`, d
     sparklines,
     tvl: sparklines.tvl[0],
     apy,
+    // echo contract state so stale hook.pricePerShare from the erc4626 hook
+    // (written before a vault turned yearn: true) gets overwritten on merge
+    pricePerShare: snapshot.pricePerShare,
     performance: {
       estimated: estimatedApr ?? undefined,
       oracle: {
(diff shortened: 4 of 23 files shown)


matheus1lva:
## Review — multi-agent pass

**Verdict: COMMENT** — 1 confirmed issue, no blockers. No linked issue; graded against the PR description.

### Issues

**1. `packages/lib/abis.ts:70` — ABI source filtering runs unconditionally, turning config drift into a silent skip**

`filterSourcesByChains(config.abis, chainIds)` runs at module load with `chainIds` derived from the always-loaded chains config — not gated on `KONG_CHAIN_IDS`. Today it is a no-op (every chainId in abis.yaml exists in chains.yaml), but under future drift (a source added to abis.yaml on a chain missing from chains.yaml) the source is silently dropped at config load. On main that misconfiguration surfaced loudly (`rpcs.next` threw `!chain, <id>` per job).

- **Change:** apply `filterSourcesByChains` only when `parseChainIds()` returns an explicit `KONG_CHAIN_IDS` override; without an override keep the previous unfiltered behavior (or warn/throw on unmatched source chains).
- **Keep:** `filterSourcesByChains` stays exported; `KONG_CHAIN_IDS=1` / `dev:ethereum` still limits static sources.
- **Done when:** with `KONG_CHAIN_IDS` unset, an abis.yaml source on an unconfigured chain is retained or fails visibly; with `KONG_CHAIN_IDS=1` filtering still works and lib specs pass.

### Suggestions (non-blocking)

1. `packages/web/app/api/gql/resolvers/strategies.ts:33` — apiVersion filter semantics changed for empty-string: old `!row.apiVersion || compare(...)` kept `''` rows; new `typeof rowApiVersion !== 'string' || compare(...)` sends `''` through `compare`, which normalizes to `'0'` and drops the row. Restore the falsy short-circuit or explicitly accept the tightening. (`vaults.ts` is behaviorally unchanged — `''` already hit compare there.)
2. `packages/lib/chains.ts` — the local-run feature (KONG_CHAIN_IDS, `dev:ethereum`, `filterSourcesByChains`) is bundled into a PPS-fix PR. Disclosed in the body; optionally split for revert granularity.
3. `packages/web/components/LatestBlocks.tsx:16` — drive-by removal of the unused map `index` param is not mentioned in the PR body.
4. `packages/ingest/abis/erc4626/timeseries/pps/hook.spec.ts` — test-style restyle (mocha `this.timeout` -> `{ timeout }` options) is not mentioned in the PR body.
5. `packages/web/lib/mergeSnapshot.ts:20` — the `asset` carve-out is the only guard against snapshot-wins shadowing hook enrichment; the invariant (hook keys must not collide with zero-arg contract views unless carved out) lives only in a comment. Worth noting where snapshot hooks are documented.
6. `packages/web/lib/mergeSnapshot.ts:3` — `SnapshotRow` duplicates `VaultSnapshot` (`rest/snapshot/db.ts`) character-for-character; `list/db.ts` imports both. Alias one to the other.

### Coverage this review did not reach

- `KONG_CHAIN_IDS` blast radius beyond ingest: the filter lives at module load in `lib/chains.ts`, so web and terminal are affected too — an invalid value throws at import time in the web app; a leaked value in a shared env silently shrinks web's chain list.
- No exhaustive audit intersecting every snapshot hook's return keys with every ABI's zero-arg views; the `asset`-only carve-out was reasoned from the pricePerShare case (spot checks found no other current collision).
- No write-side fix shipped: `erc4626/snapshot/hook.ts` still writes `pricePerShare` into the hook blob, so the stale-key mechanism keeps being created; direct hook readers still see it. The read-layer merge is the only defense.
- The accountant resolver path (a named fix target) has no end-to-end test; the containers fixture covers vault/strategy only.
- The apiVersion filter tweaks are untested, and the number-vs-string serialization claim is strictly asserted only on the REST snapshot check.

---

## How This Was Reviewed

matheus1lva:
Outcome of the review above — addressed in 2206a64:

- **Issue 1 (abis.ts:70): fixed.** `filterSourcesByChains` now runs only when `parseChainIds()` returns an explicit `KONG_CHAIN_IDS` override. Without the override, `config.abis` is used unfiltered (previous behavior), so chains.yaml/abis.yaml drift fails loudly at runtime again instead of a silent skip. `KONG_CHAIN_IDS=1` / `dev:ethereum` still limits static sources. Lib specs pass (35/35).
- **Suggestion 1 (strategies.ts:33): fixed.** Restored the empty-string short-circuit — `typeof rowApiVersion !== 'string' || !rowApiVersion || compare(...)` — so `''` apiVersion rows are kept as before this PR. Web specs pass (16/16).
- Suggestions 2–4 (bundling, LatestBlocks and test-restyle disclosure): PR-body/scoping notes, left as-is.
- Suggestions 5–6 (carve-out invariant doc, SnapshotRow/VaultSnapshot dedup) and the coverage notes: not addressed here; candidates for follow-up.

matheus1lva:
## Review — fix(web): serve contract state over stale hook keys

**Verdict: COMMENT** — one confirmed issue; the core merge fix is sound and well-covered.

The `mergeSnapshot` merge order (defaults < hook < snapshot, `asset` hook-won when non-null) is correct for the stale-key problem, the single call site covers both REST endpoints (the snapshot route reads the redis cache `getVaultsWithSnapshots` writes), and the unit spec covers precedence, the asset exception, and missing blobs.

### Issues

1. **The merged stale-hook containers suite ships red: 4 GraphQL tests fail by construction.** In `packages/ingest/yvusd-estimated-apr.containers.spec.ts`, the "contract state wins over stale hook keys" suite asserts fresh `pricePerShare` through GraphQL (`vault`, `vaults`, the four strategy resolvers, `accountVaults`), but the GraphQL resolvers were intentionally reverted to spreading `...row.hook` last, so the seeded stale hook value wins and all 4 assertions fail (suite is 5/9, disclosed in the test plan). The suite's header comment also still claims the fix covers "both REST and GraphQL".
   - Change: trim the suite to REST scope — keep the 2 REST tests and the vault fixture, drop the 4 GraphQL tests plus the strategy/account fixture rows they exist for, and update the header comment to say the merge fix is REST-only here.
   - Keep: the 3 estimated-apr tests and the 2 REST stale-hook tests, untouched.
   - Done when: `bun --filter ingest test:containers -- yvusd-estimated-apr.containers.spec.ts` is green.

### Suggestions

- `SnapshotRow` exported from `packages/web/lib/mergeSnapshot.ts` is unused anywhere — dead after the GraphQL revert; drop it.
- Coverage not reached: the test plan records runs for `mergeSnapshot.spec.ts` and the containers suite, but not for the new/changed ingest unit specs (`abis/yearn/3/vault/timeseries/pps/hook.spec.ts`, `extract/snapshot.spec.ts`, `abiutil.spec.ts`, restyled `abis/erc4626/timeseries/pps/hook.spec.ts`). An independent run during this review had the mocked specs green (8/11) with the 3 failures all live-RPC env errors (cloudflare-eth internal error, polygon-rpc 401), so the spec logic holds — but recording a run with working RPC creds in the plan would close the loop.

---

## How This Was Reviewed
5 review lenses, each finding independently verified by claude. 4 candidate findings were refuted and dropped.

matheus1lva:
## Review outcomes

- **Issue 1 (red containers suite — 4 failing GraphQL stale-hook tests):** fixed in 5f83ae5. Trimmed the "contract state wins over stale hook keys" suite to REST scope: dropped the 4 GraphQL tests and the strategy/account fixture rows that existed only for them, and updated the suite header comment to state the merge fix is REST-only (GraphQL resolvers keep the old merge). The 3 estimated-apr tests and the 2 REST stale-hook tests are untouched.
- **Suggestion (unused `SnapshotRow` export in `packages/web/lib/mergeSnapshot.ts`):** fixed in d70201a — removed.
- **Suggestion (record runs for the new ingest unit specs):** left as-is — the mocked specs were verified green during the review; the live-RPC ones only failed on public-endpoint env errors, so this is a test-plan bookkeeping item for the author.

**Test results after the fix:**
- `bun --filter ingest test:containers -- yvusd-estimated-apr.containers.spec.ts`: 5/5 passed (3 estimated-apr + 2 REST stale-hook).
- `bun --filter web test`: 16/16 passed (4 files, incl. `mergeSnapshot.spec.ts`).
- `bun --filter ingest lint`: 0 errors (pre-existing warnings only).

murderteeth (approved):
## Review — fix(web): serve contract state over stale hook keys

**Verdict: APPROVE** — no blockers.

No linked issue, so this is graded against the PR description.

`mergeSnapshot`'s order (defaults < hook < snapshot, with `asset` hook-won when non-null) is the right shape for the stale-key problem. The unit spec pins precedence, the asset exception, and missing blobs; the ten call sites (`rest/list/db.ts` plus the nine resolvers) apply it consistently; and the v3 snapshot-hook echo is a sound at-rest cleanup so vault rows self-heal on the next snapshot pass without a migration.

### One non-blocking note

`packages/ingest/yvusd-estimated-apr.containers.spec.ts:352` — the REST-list test (line 352) and the cached-list test (line 383) assert on SQL-projected list columns, not on the merged snapshot. `getVaultsWithSnapshots` builds the list item from `listColumns` (`packages/web/app/api/rest/list/db.ts:235-236`), where `pricePerShare` comes from `(snapshot.snapshot->>'pricePerShare')::numeric` (line 215) and `asset` from `snapshot.hook->'asset'` (line 154) — both byte-identical on `origin/main`. `mergeSnapshot` is applied only to the separate `snapshot` field (line 242), which never feeds `listItem`. So the seeded `hook.pricePerShare = 1002336` could not shadow those values before or after the fix, and both tests pass on unfixed code.

They're still useful guards on the list projection — just retitle them so they don't claim to cover the stale-hook fix, or add an assertion on a list-path field that actually flows through the merge. The three discriminating tests (REST snapshot line 342, GraphQL vault line 362, cached snapshot line 370) do fail under the old merge order, so the fix itself is genuinely covered.

---

## How This Was Reviewed

PR #455, opened 2026-08-03
fix(ingest): ingest legacy 8-arg v2 StrategyReported events

### LINKED ISSUE #450: Legacy 8-arg StrategyReported events are never ingested, leaving pre-0.3.2 vault APY null
## Impact

Pre-0.3.2 V2 vaults emit an 8-argument `StrategyReported` (no `debtPaid`):

```
StrategyReported(address indexed strategy, uint256 gain, uint256 loss, uint256 totalGain, uint256 totalLoss, uint256 totalDebt, uint256 debtAdded, uint256 debtRatio)
```

selector `0x2fb611faf48b1d1b91edbba34cee10c6357adee410540e4a8f7a82b6b38673e4`. Kong only knows the modern 9-argument variant (with `debtPaid`, selector `0x67f96d28...`), so legacy harvests are never ingested. Every harvest-derived feature is empty for these vaults:

- `getFirstTwoHarvestBlocks` finds < 2 harvests → `getInceptionBlockNumber` returns undefined → the `apy-bwd-delta-pps` pipeline yields nothing → `net`, `grossApr`, `weeklyNet`, `monthlyNet`, `inceptionNet` are all null. Live examples: `GET /api/rest/snapshot/1/0x19D3364A399d251E894aC732651be8B0E4e85001` (DAI 0.3.0) and `/1/0x0e8A7717A4FD7694682E7005957dD5d7598bF14A` both return `apy: { net: null, grossApr: null, ... }`.
- Strategy `lastReportDetail` / harvest-APR hooks keyed off `StrategyReported` never fire for these vaults' strategies.

## Evidence

- On-chain: the DAI 0.3.0 vault `0x19D3364A399d251E894aC732651be8B0E4e85001` has **9,907** logs with the legacy 8-arg selector in blocks 12.5M–14.5M alone.
- Production `evmlog`: **zero** `StrategyReported` rows for that vault (any signature).

This is the same class of gap as #446/#447 (which handled the 4-arg `StrategyAdded`): the event exists on chain, but the abi and hook topics only subscribe the modern signature.

## Where

- `packages/ingest/abis/yearn/2/vault/abi.ts` — only the 9-arg `StrategyReported` is present, so `getLogs` never requests the legacy topic.
- [`packages/ingest/abis/yearn/2/vault/event/StrategyReported/hook.ts#L13`](https://github.com/yearn/kong/blob/0eab013/packages/ingest/abis/yearn/2/vault/event/StrategyReported/hook.ts#L13) — topics list has only the 9-arg signature; the hook body must also tolerate the missing `debtPaid` arg.
- Downstream queries filter by `event_name = 'StrategyReported'` (e.g. [`apy.ts#L296`](https://github.com/yearn/kong/blob/0eab013/packages/ingest/abis/yearn/lib/apy.ts#L296)), so once the logs are ingested they are picked up without further changes.

## Proposed fix

Mirror #447's approach: add the 8-arg event to the flattened vault abi, add its selector to the `StrategyReported` hook topics, and handle the absent `debtPaid` in the hook (default `0n`). Like #447, existing `evmlog_strides` cover the historical ranges, so affected vaults need a stride reset + refanout before legacy harvests backfill.

Note for the fix: any APR math consuming these events should treat pre-0.3.0 `debtRatio` carefully — on < 0.3.0 vaults the corresponding struct field is `debtLimit` (absolute amount), and the event arg layouts differ again at 0.2.x (verify per-release signatures against the tagged `yearn-vaults` sources before wiring topics).

## Acceptance criteria

- [ ] Legacy 8-arg `StrategyReported` events are fetched, decoded, and stored in `evmlog` for pre-0.3.2 vaults after a stride reset.
- [ ] `apy-bwd-delta-pps` produces non-null `net`/`grossApr` for a legacy vault with ≥ 2 harvests (e.g. the DAI 0.3.0 vault above).
- [ ] Modern-vault (≥ 0.3.2) harvest ingestion and APY output are unchanged.
- [ ] Tests cover decoding both `StrategyReported` arities from the single vault abi (mirroring the #447 `StrategyAdded` spec).

Found while reviewing #447 (explains why the strategist-fee normalization bug is currently masked on legacy vaults). Related: #446.

matheus1lva:
### Summary

- Add the pre-0.3.2 8-argument `StrategyReported` event to the vault abi and the hook topics. The hook sets the missing `debtPaid` argument to `0n`.
- Decode v2 strategy structs with the vault api version (`strategiesAbi` + `mapStrategyParams`). Pre-0.3.2 vaults return an 8-field struct with zero-padded return data. The 9-field abi decodes that data without an error and reads `totalGain` in the `totalDebt` slot.
- Correct the strategist fee units in `extractFees__v2`. The product `performanceFee * debtRatio` has units of bps². The old code divided the sum by 10,000 one time. A legacy vault such as `0x0e8A7717A4FD7694682E7005957dD5d7598bF14A` then got a performance fee of 980.1 and a negative gross APR. The new code scales the product back to bps first. This bug predates this PR, but this PR is what exposes the vaults it breaks, so it cannot be split off.
- Set `debtRatio` to null in the reports api for 0.2.x vaults. On these vaults the event argument is `debtLimit`, an absolute token amount, not a ratio.

Closes #450.

### How to review

Start with `vault/event/StrategyReported/hook.ts`. Then follow the version-aware reads in `vault/snapshot/hook.ts`, `strategy/snapshot/hook.ts`, and `lib/apy.ts`. End with the `debtRatio` guard in `web/app/api/gql/resolvers/vaultReports.ts` and `web/app/api/rest/reports/db.ts`.

### Test plan

- [ ] `cd packages/ingest && bunx vitest run abis/yearn/2/vault/event/StrategyReported abis/yearn/2/vault/snapshot abis/yearn/2/strategy abis/yearn/lib/apy.spec.ts`
- [ ] `cd packages/ingest && npx tsc --noEmit`
- [ ] `bun --filter ingest lint`

New tests use live mainnet RPC. They lock the fee scaling on a 0.3.0 vault (0.198) and on a 0.4.3 vault with a non-zero strategist fee (0.14126). One test shows the 9-field abi misreads padded legacy return data, and shows the version-aware abi reads it correctly. One test pins the strategies abi arity that `extractFees__v2` requests per api version, because `performanceFee` and `debtRatio` sit at the same offsets in both layouts and a fee assertion alone cannot catch a revert to the hardcoded 9-field abi.

### Manual test

Prerequisites: Docker, Bun, tmux, curl, jq, and an archive-capable Alchemy key. This uses local Postgres/Redis and read-only mainnet RPC.

- [ ] Install dependencies and set the correct env vars.

- [ ] Set up the local config files.

```bash
cat > config/chains.local.yaml <<'EOF'
chains: [mainnet]
EOF

cat > config/abis.local.yaml <<'EOF'
cron: { name: AbiFanout, queue: fanout, job: abis, schedule: '*/15 * * * *', start: false }
abis:
  - abiPath: yearn/2/vault
    sources:
      - { chainId: 1, address: '0x19D3364A399d251E894aC732651be8B0E4e85001', inceptBlock: 11673762, startBlock: 11673762, endBlock: 14000000 }
  - abiPath: yearn/2/strategy
    things: { label: strategy, filter: [{ field: apiVersion, op: '<', value: '3.0.0' }] }
EOF

cat > config/manuals.local.yaml <<'EOF'
manuals:
  - chainId: 1
    address: '0x19D3364A399d251E894aC732651be8B0E4e85001'
    label: vault
    defaults: { yearn: true, origin: yearn, asset: '0x6B175474E89094C44Da98b954EedeAC495271d0F', apiVersion: '0.3.0', decimals: 18, inceptBlock: 11673762, inceptTime: 1610900779 }
EOF
```

Expected: the run only includes the legacy DAI yVault and blocks `11673762..14000000`.

- [ ] Start the standard dev environment.

```bash
make dev
```

- [ ] In the **Terminal** pane, run `ingest` → `fanout abis` → **Yes** again. In the **psql** pane, wait until `snapshot_ready` becomes `t`.

- [ ] Wait until the full ingest flow completes.

- [ ] From a second shell at the repository root, refresh the REST cache. Then run the API assertions.

```bash
bun run packages/web/app/api/rest/refresh-vaults.ts
bun run packages/web/app/api/rest/reports/refresh-historical.ts
bun run packages/web/app/api/rest/timeseries/refresh-historical.ts

curl -fsS http://localhost:3001/api/rest/snapshot/1/0x19D3364A399d251E894aC732651be8B0E4e85001 \
  | jq -e '.apiVersion == "0.3.0" and (.debts | length >= 1) and .apy.net != null and .apy.grossApr != null'

curl -fsS http://localhost:3001/api/rest/reports/1/0x19D3364A399d251E894aC732651be8B0E4e85001 \
  | jq -e 'type == "array" and length >= 2 and any(.[]; .eventName == "StrategyReported" and .debtPaidUsd != null)'

curl -fsS http://localhost:3001/api/rest/timeseries/apy-historical/1/0x19D3364A399d251E894aC732651be8B0E4e85001 \
  | jq -e 'type == "array" and any(.[]; .component == "net" and .value != null)'

curl -fsS -X POST http://localhost:3001/api/gql -H 'content-type: application/json' \
  --data '{"query":"{ vault(chainId: 1, address: \"0x19D3364A399d251E894aC732651be8B0E4e85001\") { apiVersion apy { net grossApr } debts { totalDebt } } vaultReports(chainId: 1, address: \"0x19D3364A399d251E894aC732651be8B0E4e85001\") { eventName debtPaidUsd } }"}' \
  | jq -e '((.errors // []) | length == 0) and .data.vault.apiVersion == "0.3.0" and (.data.vault.debts | length >= 1) and .data.vault.apy.net != null and .data.vault.apy.grossApr != null and (.data.vaultReports | length >= 2) and any(.data.vaultReports[]; .eventName == "StrategyReported" and .debtPaidUsd != null)'
```

Expected: each assertion prints `true` and exits with code 0. If a report, a debt, or an APY value is missing, `jq` prints `false` and exits with a non-zero code.

- [ ] Clean up. This step is local-only and destructive.

```bash
make down
rm -f .env config/chains.local.yaml config/abis.local.yaml config/manuals.local.yaml
```

Expected: this removes the local containers and the test-only config. This does not change production or chain state.

### Risk / impact

- Legacy v2 reports enter the existing report and APY flow. Modern reports keep their decode path.
- Deployment: existing `evmlog_strides` cover the legacy block ranges. Affected vaults need a stride reset and a refanout before legacy harvests backfill. This is the same procedure as #447.
- A replay changes stored values. Legacy strategy `totalDebt` changes value, because the old read was quietly misaligned. Performance fees decrease to the correct bps value on v2 vaults with a non-zero strategist fee. This corrects their gross APR.


murderteeth (changes requested):
The core of this is right. The legacy event is in the ABI and subscribed in the hook topics, `debtPaid` defaults cleanly, and the version-aware split between `strategiesAbi` and `mapStrategyParams` is the correct shape. Switching `fetchPreviousHarvest` to match either selector is the subtle part that makes harvest APR work at all for these vaults, and it's easy to miss — nice.

Two things to sort out before this lands.

**1. Legacy vaults will report a negative APR, which undoes the point of the fix.**

Once these harvests are ingested, `_compute` stops bailing early and legacy vaults reach `extractFees__v2` for the first time. That function has a units bug: it multiplies the strategy's performance fee by its debt ratio, giving basis-points-squared, then adds the vault's plain-basis-points fee and divides the total by 10,000 just once.

For `0x0e8A7717A4FD7694682E7005957dD5d7598bF14A`, the 0.3.0 vault named in the issue, that works out to a performance fee of 980.1 instead of 0.198. Since gross APR is computed as `netApr / (1 - performance)`, the sign flips and the API publishes a negative gross APR. So the vault goes from "APY is null" to "APY is wrong", which isn't the outcome #450 is asking for.

The fix is to scale the per-strategy product back down to basis points before summing it with the vault fee. Worth confirming the existing yvUSDT spec still reports 0.2 — it passes today only because that strategy's fee happens to be zero, which is why nobody has hit this.

To be clear, this bug predates your PR. But this PR is what exposes the vaults it breaks, so it can't really be split off.

**2. Neither new test would catch a regression in the thing you fixed.**

You can revert the version-aware ABI selection in `apy.ts` back to the hardcoded 9-field version and the whole suite stays green. Two reasons:

The decode guard in `hook.spec.ts` asserts that the flattened ABI *throws* on a legacy return. It doesn't, in practice — real pre-0.3.2 vaults zero-pad their return data, so the 9-field ABI decodes it happily and just reads the wrong slot, handing back `totalGain` where `totalDebt` should be. Silent misalignment is the failure mode worth pinning; a length check isn't. (The comment about multicall entries failing and fees collapsing to zero is off for the same reason — those calls succeed.)

And the test named for `extractFees__v2` doesn't call it. It calls the mapper and then repeats the fee arithmetic inside its own assertion, so it can't fail when the function changes. The only test that does call `extractFees__v2` passes `'0.4.3'`, which never reaches the new branch.

**Smaller things:**

The PR says no migration or deployment step is introduced, but the issue's first acceptance criterion depends on a stride reset and refanout — existing strides already cover these ranges, so nothing backfills on its own. #447 called this out in its commit message. Also worth mentioning that legacy strategy `totalDebt` will *change value* on replay, since the old read was quietly misaligned.

Finally, on 0.2.x vaults the trailing event arg is `debtLimit`, an absolute amount, not a ratio. Your comment says so, but the reports API still serves it under `debtRatio` with no version guard, so consumers can't tell the difference.

---

## How This Was Reviewed
5 review lenses, each finding independently verified by codex. 0 candidate findings were refuted and dropped.

murderteeth (approved):

PR #457, opened 2026-08-06
Add bounded batch PPS endpoint

w84april:
### Summary

Add a Redis-backed endpoint for fetching bounded PPS histories across multiple chains. This lets consumers request one shared date range for several vaults without downloading each vault’s complete timeline.

- Validate shared date ranges and chain-qualified vault addresses
- Preserve recent-over-historical precedence and predecessor anchors
- Bound request and response sizes and document the contract

### How to review

Review the implementation in this order:

1. `contract.ts` — request validation, address normalization, and limits
2. `service.ts` — Redis reads, historical/recent merging, and range selection
3. `route.ts` — HTTP handling, CORS, and payload-size protection
4. `docs/rest.md` — public request and response contract

The existing REST timeseries endpoint, GraphQL API, ingestion flow, and database schema are intentionally unchanged.

For a local smoke test:

```bash
curl -s http://localhost:3001/api/rest/timeseries/pps/v2 \
  -H 'Content-Type: application/json' \
  -d '{
    "start": 1785628800,
    "finish": 1785715200,
    "addresses": [
      "1:0x0000000000000000000000000000000000000000",
      "137:0x0123000000000000000000000000000000000123"
    ]
  }' | jq

### DIFF
--- a/packages/web/app/api/rest/timeseries/pps/v2/contract.ts
+++ b/packages/web/app/api/rest/timeseries/pps/v2/contract.ts
@@ -1,0 +1,133 @@
+const DAY_SECONDS = 86_400
+const MAX_ADDRESSES = 50
+const MAX_RANGE_YEARS = 10
+const MAX_CHAIN_ID = 2_147_483_647
+const ADDRESS_PATTERN = /^0x[0-9a-fA-F]{40}$/
+const QUALIFIED_ADDRESS_PATTERN = /^([1-9][0-9]*):(0x[0-9a-fA-F]{40})$/
+const REQUEST_FIELDS = new Set(['start', 'finish', 'addresses'])
+
+export type PpsAddress = {
+  key: string
+  chainId: number
+  address: string
+}
+
+export type PpsBatchRequest = {
+  start: number
+  finish: number
+  addresses: PpsAddress[]
+}
+
+export type PpsPoint = {
+  time: number
+  value: string
+}
+
+export type PpsBatchResponse = Record<string, PpsPoint[]>
+
+export type ParsePpsBatchRequestResult =
+  | { success: true; data: PpsBatchRequest }
+  | { success: false; error: string }
+
+function isRecord(value: unknown): value is Record<string, unknown> {
+  return typeof value === 'object' && value !== null && !Array.isArray(value)
+}
+
+function isUtcDayTimestamp(value: unknown): value is number {
+  return typeof value === 'number' && Number.isSafeInteger(value) && value >= 0 && value % DAY_SECONDS === 0
+}
+
+function isWithinRangeLimit(start: number, finish: number): boolean {
+  const startDate = new Date(start * 1_000)
+  if (!Number.isFinite(startDate.getTime())) return false
+
+  const targetYear = startDate.getUTCFullYear() + MAX_RANGE_YEARS
+  const targetMonth = startDate.getUTCMonth()
+  const lastTargetDay = new Date(Date.UTC(targetYear, targetMonth + 1, 0)).getUTCDate()
+  const targetDay = Math.min(startDate.getUTCDate(), lastTargetDay)
+  const maximumFinish = Date.UTC(targetYear, targetMonth, targetDay) / 1_000
+
+  return finish <= maximumFinish
+}
+
+function parseQualifiedAddress(value: unknown): PpsAddress | undefined {
+  if (typeof value !== 'string') return undefined
+
+  const match = QUALIFIED_ADDRESS_PATTERN.exec(value)
+  if (!match) return undefined
+
+  const chainId = Number(match[1])
+  const address = match[2].toLowerCase()
+
+  if (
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: packages/web/app/api/rest/timeseries/pps/v2/contract.ts [hunk 1]
```diff
@@ -0,0 +1,133 @@
...
+  if (!match) return undefined
+
+  const chainId = Number(match[1])
+  const address = match[2].toLowerCase()
+
+  if (
+    !Number.isSafeInteger(chainId) ||
+    chainId <= 0 ||
+    chainId > MAX_CHAIN_ID ||
+    !ADDRESS_PATTERN.test(address)
```
matheus1lva, line 66: https://viem.sh/docs/utilities/getAddress can help with this

packages/web/app/api/rest/timeseries/pps/v2/contract.ts
```diff
@@ -0,0 +1,133 @@
...
+  }
+
+  return {
+    key: `${chainId}:${address}`,
+    chainId,
+    address,
+  }
+}
+
+export function parsePpsBatchRequest(input: unknown): ParsePpsBatchRequestResult {
```
matheus1lva, line 78: wdyt of using zod here? will make your life easier!

packages/web/app/api/rest/timeseries/pps/v2/route.ts
```diff
@@ -0,0 +1,70 @@
...
+}
+
+export async function POST(request: Request) {
+  let body: unknown
+
+  try {
+    const contentLength = Number(request.headers.get('content-length'))
+    if (Number.isFinite(contentLength) && contentLength > MAX_REQUEST_BYTES) {
+      return errorResponse('Request body is too large', 413)
+    }
```
matheus1lva, line 30: maybe next already handles that for us? I dont remember if does.

matheus1lva (commented):
No linked issue; graded against the PR description. All described behaviors (validation, recent-over-historical precedence with predecessor anchor, size bounds, docs) are implemented.

## Issues

**1. One malformed cached row 500s the whole batch** — `service.ts:46`
`toPpsCandidate`/`toPpsPoint` throw on a bad row inside `mergeAndSlicePps`; `route.ts` catches and returns 500 for all up-to-50 requested series. v1 tolerates the same data by filtering. If per-series blast radius matters, catch per address in `readPpsBatch` and return an empty array for the corrupt series; keep precedence and anchor semantics unchanged. Done when a batch with one malformed cached row still returns 200 with the valid series intact — or confirm fail-whole-batch is the intended contract (docs do document 500 on cache read failure).

**2. 16KB request cap bypassable without content-length** — `route.ts:27`
`Number(request.headers.get('content-length'))` is `0` for chunked/absent header, so the guard passes; `request.text()` then buffers the full body before the byte check. Stream with an incremental cap (abort past `MAX_REQUEST_BYTES`) or reject requests lacking a valid content-length. Done when an oversized chunked POST gets 413 without full buffering; valid <=16KB bodies still return 200.

**3. Second Redis client instead of the shared one** — `service.ts:12`
Module-level `createKeyv(process.env.REST_CACHE_REDIS_URL || ...)` duplicates `cache.ts`'s `getKeyvClient()` config; every other REST route reuses the shared client. Reuse it — but v2 relies on `{ throwOnErrors: true }` for its 500 path, which the shared client doesn't set, so export a throwing variant deliberately rather than silently dropping error propagation. Keep the injectable `cacheGet` seam. Done when no `createKeyv` call remains under `pps/v2` and Redis read failures still 500.

**4. No tests despite explicit test seams** — `service.ts:66`
`mergeAndSlicePps` is exported and `readPpsBatch` takes injectable `cacheGet`, yet the PR ships zero specs; the package has vitest and the `.spec.ts` convention next door (`cache.spec.ts`). Add `service.spec.ts` + `contract.spec.ts` covering anchor at/before start, recent-over-historical tie-break, empty series, address dedup/normalization, range validation, and key ordering via stubbed `cacheGet`. Done when `bun --filter web test` exercises all three and passes.

## Suggestions (non-blocking)

- `service.ts:26` — cached-row validation is split between `toPpsCandidate` and `toPpsPoint`; consolidating it in one place would make the throw paths obvious.
- Coverage this review did not reach: the producer-to-consumer cache contract (whether `refresh.ts`/`refresh-historical.ts` writer output actually satisfies the new strict validators — pg numeric strings, day alignment); the `cache-control: no-store` divergence from v1's 900s public caching; the 4MB response cap firing only after the full Redis read + stringify; and docs/rest.md accuracy vs implementation (405 behavior, address dedup, exact-start anchor, no-store are undocumented).

---

## How This Was Reviewed

murderteeth (changes requested):
Blocking this one for now. As designed it creates a griefing vector. Lets discuss on telegram

Valid request parameters are effectively unlimited and cost the same regardless of range, making it possible to grief redis

PR #461, opened 2026-08-27
perf(prices): batch price service lookups and extend block cache ttl

matheus1lva:
### Summary
The indexer hit `/api/prices/historical` once per token per block, and cached each answer for 30 seconds.
Since lib/blocks pins the head block for 15 minutes and the price service refreshes today's row hourly,
that TTL re-fetched an unchanged value roughly 120 times an hour per token. This coalesces concurrent
lookups into one `batchHistorical` call and holds a service hit for 15 minutes.

### How to review
Start with `packages/ingest/prices.ts`:

- `enqueuePriceServiceBatch` / `flushPriceServiceBatch` / `sendPriceServiceBatch` — the batching layer.
  Entries are keyed by `coinId:utcDay`, flushed after 25ms or at 50 queued entries, whichever comes first.
  Note that `sendPriceServiceBatch` never throws: on any failure it resolves every entry as `found: false`.
- `fetchPriceServiceExactResult` — the prior per-token path, extracted unchanged. `batchHistorical` reads
  the service's table only, so a batch miss still falls through here, where the service resolves upstream.
- `blockCacheTtl()` — returns a function under `USE_PRICE_SERVICE` so hits get the 15m TTL while misses
  keep the short negative TTL. A transient outage must not stick `tvl=0` for 15 minutes.

Two behaviors worth a second look: a stored zero comes back as `missing` without an exact-endpoint call
(the exact route would read the same row), and batch response keys are lowercased on read because the
service echoes the checksummed address.

### Test plan
- [ ] Manual: run the indexer with `USE_PRICE_SERVICE=true` against a chain with several vaults and
      confirm one `batchHistorical` request per flush window instead of one `historical` request per token.
- [ ] Automated (when applicable): `bun --filter ingest test` — not run.

### Risk / impact
Priced data feeds TVL, so a wrong or missing price shows up in user-facing numbers. Guards against that:
batch failures degrade to the existing exact endpoint rather than a miss, and negative results keep the
short TTL. `USE_PRICE_SERVICE` is the kill switch — off, `blockCacheTtl()` returns the old 30s constant
and none of the batching code runs.

### DIFF
--- a/packages/ingest/prices.service-cache.mock.spec.ts
+++ b/packages/ingest/prices.service-cache.mock.spec.ts
@@ -1,7 +1,8 @@
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
 
-const { cacheGet, cacheSet, blockTime, wrapStore } = vi.hoisted(() => ({
-  cacheGet: vi.fn(), cacheSet: vi.fn(), blockTime: vi.fn(), wrapStore: new Map<string, unknown>()
+const { cacheGet, cacheSet, blockTime, wrapStore, wrapTtls } = vi.hoisted(() => ({
+  cacheGet: vi.fn(), cacheSet: vi.fn(), blockTime: vi.fn(),
+  wrapStore: new Map<string, unknown>(), wrapTtls: [] as unknown[]
 }))
 
 vi.mock('lib/blocks', () => ({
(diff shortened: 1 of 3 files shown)


murderteeth (approved):
## Summary
Coalesces per-token `/api/prices/historical` lookups into a single `batchHistorical` call behind a 25ms / 50-entry flush window, and replaces the flat 30s block-cache TTL with a result-dependent TTL under `USE_PRICE_SERVICE` — 15m for a service hit, 120s for a miss. A batch miss or batch failure falls through to the extracted `fetchPriceServiceExactResult`, so no path loses its upstream resolution.

## Issues
None.

## Verdict
APPROVE

---

## How This Was Reviewed
5 review lenses, each finding independently verified by claude. 3 candidate findings were refuted and dropped.

PR #462, opened 2026-08-27
fix(ingest): stop timeseries fanout loop on unavailable prices

matheus1lva:
### Summary
Since #442, a day where the price service errors writes no output row (`tvl.ts` returned `[]`
on `priceSource === 'unavailable'`). The timeseries fanout treats a missing row as "not yet
computed", so it re-enqueued those days every cycle, forever. The fanout query also returned
each vault's entire computed-day history per cycle: ~50M calls and 745M rows in
`pg_stat_statements`, which pushed Neon egress from ~40 to ~230GB/day starting Aug 22.

Two changes:
- `fanout/timeseries.ts`: new `findMissingDays` anti-joins the day timeline against `output`
  and returns only the gaps, keeping the single index-only range scan on
  `idx_output_chain_address_label_series_time`.
- `abis/yearn/lib/tvl.ts`: unavailable prices on past days now write rows with null usd
  components (`tvl`, `delegated`, `priceUsd`) and real on-chain values (`totalAssets`,
  `delegatedAssets`). The row closes the day and ends the loop. The current day still skips
  the write: fanout re-extracts it every cycle anyway, and a null row at today's
  `series_time` would shadow yesterday's real value in latest-row queries.

### How to review
Start with `findMissingDays` in `packages/ingest/fanout/timeseries.ts`: the `NOT IN`
subquery must keep the `BETWEEN` range so Timescale still prunes chunks. Then the
`unavailable` branch in `packages/ingest/abis/yearn/lib/tvl.ts`. The spec
`fanout/timeseries.spec.ts` pins the subtle part: `value: null` survives zod parse, the
Redis JSON roundtrip, and the upsert (writes NULL, replay overwrites it later).

### Test plan
- [ ] Manual: after deploy, watch Neon egress and the `timeseries-*` queue depth drop once
      the null rows land
- [x] Automated: `vitest run --project mocks tvl.unavailable` (4 tests),
      `vitest run --project ingest fanout/timeseries` (4 tests, testcontainers)

### Risk / impact
Past days nulled during a transient service blip stay null until someone replays them; the
old behavior retried forever but self-healed. The current day self-heals via the
unconditional end-of-timeline re-extract. `apy.ts` keeps its own permanent `return []`
paths (`!vault`, `!apy`), so those days remain immortal, but the fanout fix caps their cost
to near-zero rows per cycle. No migration; rollback is a plain revert.

### DIFF
--- a/packages/ingest/abis/yearn/lib/tvl.ts
+++ b/packages/ingest/abis/yearn/lib/tvl.ts
@@ -34,16 +34,24 @@ export default async function _process(chainId: number, address: `0x${string}`, 
 
   // extractTotalAssets returns undefined on multicall failure; skip emitting a false zero (a genuine empty vault is 0n)
   if (totalAssets === undefined) return []
-  if (priceSource === 'unavailable') return []
+
+  // 'unavailable' (transient price service failure) still writes rows for past days:
+  // null for the usd components, real values for the on-chain ones. Returning []
+  // left the day permanently missing, so every fanout cycle re-enqueued it forever
+  // (fanout/timeseries.ts). Null days heal only by replay. The current day is the
+  // exception: fanout re-extracts it every cycle anyway, and a null row at today's
+  // series_time would shadow yesterday's real value in latest-row queries.
+  const unavailable = priceSource === 'unavailable'
+  if (unavailable && latest) return []
 
   if (components) {
     // componentized outputs
     return OutputSchema.array().parse([{
       chainId, address, blockNumber, blockTime: data.blockTime, label: data.outputLabel,
-      component: 'tvl', value: tvl
+      component: 'tvl', value: unavailable ? null : tvl
     }, {
       chainId, address, blockNumber, blockTime: data.blockTime, label: data.outputLabel,
-      component: 'delegated', value: delegatedTvl
+      component: 'delegated', value: unavailable ? null : delegatedTvl
     }, {
       chainId, address, blockNumber, blockTime: data.blockTime, label: data.outputLabel,
       component: 'totalAssets', value: normalize(totalAssets, decimals) || 0
@@ -52,14 +60,14 @@ export default async function _process(chainId: number, address: `0x${string}`, 
       component: 'delegatedAssets', value: normalize(delegatedAssets, decimals) || 0
     }, {
       chainId, address, blockNumber, blockTime: data.blockTime, label: data.outputLabel,
-      component: 'priceUsd', value: priceUsd
+      component: 'priceUsd', value: unavailable ? null : priceUsd
     }])
 
   } else {
     // legacy tvl output
     return OutputSchema.array().parse([{
       chainId, address, blockNumber, blockTime: data.blockTime, label: data.outputLabel,
-      component: 'tvl', value: tvl
+      component: 'tvl', value: unavailable ? null : tvl
     }])
 
   }
--- a/packages/ingest/abis/yearn/lib/tvl.ts
+++ b/packages/ingest/abis/yearn/lib/tvl.ts
@@ -8,7 +8,8 @@ import { normalize, priced } from 'lib/math'
 import { extractWithdrawalQueue } from '../2/vault/snapshot/hook'
 import { Data } from '../../../extract/timeseries'
 import { estimateHeight, getBlock } from 'lib/blocks'
-import { first } from '../../../db'
+import { first, some } from '../../../db'
+import { endOfDay } from 'lib/dates'
 
 export default async function _process(chainId: number, address: `0x${string}`, data: Data, components?: boolean): Promise<Output[]> {
   console.info('🧮', data.outputLabel, chainId, address, (new Date(Number(data.blockTime) * 1000)).toDateString())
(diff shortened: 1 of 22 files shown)


murderteeth (commented):
## Summary
The anti-join in `findMissingDays` is the right shape for the egress problem, and pinning `value: null` through zod → Redis → upsert in the spec is the right thing to pin.

Scope note: this branch also carries the price-service batching layer from #461 (`prices.ts`, +122/−14). That code was reviewed on #461 and is not re-graded here, so the finding below is the only one covering this diff.

## Issues

- [ ] **Outage days publish a false $0 TVL point (high)** — `packages/ingest/abis/yearn/lib/tvl.ts:51` — a day nulled by a price-service blip now appears in the day-granularity timeseries as a real $0 TVL instead of being absent from the series, so charts show a crash to zero that never happened. Worse than before: `findMissingDays` counts the null row as computed, so the day is never re-enqueued and the $0 stays until someone manually replays it. The old `return []` was noisy but self-healed.
  - Done when: a `tvl-c`/`tvl` row whose `value` is NULL is either omitted from the gql `timeseries`/`tvls` and REST timeseries responses at 1-day granularity, or carries something consumers can tell apart from a genuine zero — with a test reading a NULL-valued row back through one of those queries.
  - Provenance: `0673e85`

## Verdict
COMMENT — no confirmed blocker; one confirmed issue.

---

## How This Was Reviewed
5 review lenses, each finding independently verified by claude. 5 candidate findings were refuted and dropped.

matheus1lva:
Addressed the outage-day $0 finding.

**SQL side** (`4e27908`, was unpushed when you reviewed): all-null day buckets are dropped rather than reported as 0 —
- `packages/web/app/api/gql/resolvers/timeseries.ts` (both `alltimeseries` and `yearntimeseries`)
- `packages/web/app/api/gql/resolvers/tvls.ts`
- `packages/web/app/api/rest/timeseries/db.ts` (`getFullTimeseries`, `getRecentTimeseries`)

all now carry `HAVING COUNT(value) > 0`. Same commit makes the `tvl-gaps` detector's `minTvl` prefilter read the last
**non-null** tvl, so a nulled day no longer hides a vault from gap detection, and stops the tvl hook from overwriting a
day that already holds a real tvl (`hasComputedTvl`).

**REST route** (`cd706b9`, new): the SQL guard doesn't cover points already cached in Redis from before it. The route
served those straight through as `value: null`, which reads as $0 downstream. It now omits null-valued points.

**Test:** `packages/web/app/api/rest/timeseries/[segment]/[chainId]/[address]/route.spec.ts` reads a NULL-valued row
back through the REST timeseries endpoint and asserts it is absent from the response — both for a null historical day
and for a null latest row overriding a real historical one.

CI: lint + test green.

murderteeth (changes requested):
## Summary
The `HAVING COUNT(value) > 0` guards and the REST null filter close the day-granularity path from the last review. One read path was missed, and it's the one that feeds the vault's headline number.

## Issues

- [ ] **Vault headline TVL still reads $0 during an outage (high)** — `packages/ingest/db.ts:92` — the new `FILTER (WHERE value IS NOT NULL)` fixes a bucket that mixes null and real rows, but a 7-day bucket whose rows are *all* null is still emitted, and `COALESCE(..., 0)` turns it into `close: 0`. That point becomes `snapshot.hook.tvl` → gql `vault.tvl` and the vault-list sort key (`resolvers/vaults.ts:36`), so the vault shows $0 and sorts last. Permanent for a chain absent from `PRICE_SERVICE_CHAIN_NAMES` or an unset `PRICE_SERVICE_API_KEY`. The four other read paths this commit touched drop the all-null bucket; this one does not.
  - Done when: with `output` holding only null-valued `tvl-c`/`tvl` rows in the newest 7-day bucket, `getSparkline`'s newest entry is absent rather than `close: 0` — while a mixed bucket still closes on its last real value and a genuine zero still closes at 0. Covered by a test.
  - Provenance: `4e27908`

- [ ] **Replay silently drops a stored price for empty vaults (medium)** — `packages/ingest/abis/yearn/lib/tvl.ts:54` — the `hasComputedTvl` guard is gated on `usdUnknown`, which is false when `totalAssets === 0n`, while the `priceUsd` write on line 72 is gated on `priceUnavailable` alone. Re-extracting a past zero-asset day while the service is down overwrites a real stored `priceUsd` with null. Not outage-only: `unavailable` is also returned permanently for an unmapped chain or a missing API key.
  - Done when: re-running the hook for a past day with `priceSource: 'unavailable'` and `totalAssets` of `0n`, against a row already holding a non-null `priceUsd`, leaves that value unchanged — with a test covering the empty-vault case.
  - Provenance: `4e27908`

- [ ] **The five new SQL guards have no test (medium)** — `packages/web/app/api/gql/resolvers/timeseries.ts:57` — `HAVING COUNT(value) > 0` in gql `alltimeseries`/`yearntimeseries`, gql `tvls`, and REST `getFullTimeseries`/`getRecentTimeseries` is the only thing keeping outage days out of those responses, and nothing fails if one is removed. `route.spec.ts` pins only the post-cache JS filter, which cannot backstop them — an unguarded bucket returns `0`, not `null`. `tvls.spec.ts` already asserts on emitted SQL and runs in CI.
  - Done when: each of the five aggregate read paths has a test that fails if that path stops excluding an all-null 1-day bucket.
  - Provenance: `4e27908`

## Verdict
REQUEST_CHANGES — one confirmed blocker.

---

## How This Was Reviewed

murderteeth (approved):
## Summary
Both fixes land where the findings pointed: `HAVING COUNT(value) > 0` in `getSparkline` drops the all-null bucket instead of coalescing it to `close: 0`, and re-gating the `hasComputedTvl` early return on `priceUnavailable` rather than `usdUnknown` covers the zero-asset replay case. The five aggregate guards are now pinned by SQL assertions in CI-covered specs.

## Issues
None.

## Verdict
APPROVE.

---

## How This Was Reviewed

PR #463, opened 2026-08-30
fix(ingest): exponential backoff on price-service negative cache

matheus1lva:
## What

Retry loop against prices.yearn.dev: kong re-requests the same failing historical prices every ~2 minutes, each attempt costing ~14 metered drpc calls (~1.5M/day vs ~700 baseline). #462 bounded the loop; this stops the spin.

Root cause is a cache-ttl asymmetry in `packages/ingest/prices.ts`: a successful past-day price caches for 24h, a failed one for 2m. With the service failing ~51% of requests, every failure returned to the queue almost immediately — 720 attempts/day per failing key.

## How

- Repeat failures on the same day key double the negative-cache ttl: 2m → 4m → 8m → … capped at 6h (~13 attempts/day worst case).
- First failure still retries in 2m, so a transient outage can't stick tvl=0 for hours.
- Attempt counter lives under a separate key with a 24h ttl so it survives the negative marker's expiry.
- A replay fanout (`fanout replays`) now flushes the price-service negative markers and attempt counters from Redis before enqueuing, so an escalated 6h marker can't turn the replay into a no-op. Cached real prices are kept.
- `packages/lib/cache.ts` gains a `keys(pattern)` accessor for that flush, backed by node-redis `scanIterator` (non-blocking SCAN, not KEYS); deletes run in batches of 100.

## Tests

- Updated `prices.service-cache.mock.spec.ts` for the extra attempts-key write.
- New test: repeat failures escalate the ttl 2m/4m/8m and cap at 6h.
- New test: replay clear removes negative markers and attempt counters, keeps cached prices.
- `bun --filter ingest test`: 179 passed. `bun --filter lib test`: 29 passed.

### DIFF
--- a/packages/ingest/fanout/abis.ts
+++ b/packages/ingest/fanout/abis.ts
@@ -1,5 +1,6 @@
 import { abisConfig, chains, mq, sentry } from 'lib'
 import * as things from '../things'
+import { clearNegativePriceCache } from '../prices'
 import WebhookCollector from './webhooks'
 import { findBusyMatch } from './isBusy'
 
@@ -15,6 +16,8 @@ export default class AbisFanout {
       })
       return
     }
+
+    if ((data as { replay?: { enabled?: boolean } }).replay?.enabled) await clearNegativePriceCache()
 
     const webhookCollector = new WebhookCollector()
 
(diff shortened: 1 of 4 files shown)


murderteeth (approved):
## Review: `APPROVE`

The backoff is correctly scoped to the past-day day-key branch, the attempts counter is deliberately kept outside the negative marker's TTL, and the replay flush uses `SCAN` rather than `KEYS`. Five lenses produced five candidate defects; every one was refuted on the code.

No confirmed defects.

### Checks

- Lint 0 errors. The PR's own spec passes 12/12. The 27 other failures across ingest+lib are `cloudflare-eth.com` RPC calls with no egress in the review sandbox, in untouched specs.
- Visual verification — skipped, no UI surface.

PR #466, opened 2026-09-01
fix(ingest): retry price-service 5xx and stop genesis BlockNotFound spam

matheus1lva:
### Summary
Price-service 5xx/timeouts were either stampeded onto `/historical` per token or persisted as a fake USD 0. `estimateHeight` also hard-required block 1, which pruned RPCs throw as BlockNotFound (~800/h). This PR retries transport/5xx with a 5s timeout (404 stays final), treats a failed batch as unavailable instead of fanning out to exact, lets **null** tvl/tvl-c USD rows heal without looping apy/pps (#462), and bounds interpolation to the earliest reachable block (cached).

### Related
- Service-side complement: [yearn/yearn-prices#51](https://github.com/yearn/yearn-prices/pull/51) — look up requested UTC EOD in `token_prices`; miss → DeFiLlama; persist today's EOD so later same-day requests are table hits (stops Llama 429 stampede). Does **not** serve yesterday's close for the open day.

### How to review
- `packages/ingest/prices.ts`: `fetchPriceServiceResponse` timeout+retry; batch 5xx → `unavailable` (no exact stampede); batch 404 still falls through to exact; exact 5xx → unavailable, 404 → na.
- `packages/ingest/fanout/timeseries.ts`: `findMissingDays` treats null **tvl** component as missing only for labels `tvl`/`tvl-c`. Real 0 stays computed. Other labels keep “any row closes the day”.
- `packages/lib/blocks.ts`: do not require genesis/1; swallow those misses; cache `earliestReachableBlock`.

### Test plan
- [x] `bunx vitest run --project mocks prices.service.mock.spec.ts prices.service-cache.mock.spec.ts` — 27/27 passed
- [ ] ingest `fanout/timeseries.spec.ts` and lib `blocks.spec.ts` need Docker/testcontainers (not available in this environment)

### Risk / impact
Null-USD days re-enter the extract queue. jobId dedupe prevents in-flight stacks; `#463` negative cache (2m→6h) limits price-service load. `removeOnComplete` count is 100, so jobIds recycle and extract can re-run ~each fanout cycle during a long outage (RPC for totalAssets + estimateHeight). That is not the #462 Neon anti-join egress loop. Non-tvl labels are unchanged so apy/pps cannot loop. 5xx is never stored as a real 0; na/404 still is.

### DIFF
--- a/packages/ingest/abis/yearn/lib/tvl.ts
+++ b/packages/ingest/abis/yearn/lib/tvl.ts
@@ -38,8 +38,8 @@ export default async function _process(chainId: number, address: `0x${string}`, 
 
   // 'unavailable' (price service failure) still writes rows for past days: null for
   // the usd components, real values for the on-chain ones. Returning [] left the day
-  // permanently missing, so every fanout cycle re-enqueued it forever. Null days heal
-  // by replay (fanout/timeseries.ts).
+  // with no row; fanout would re-enqueue every cycle during a long outage. Null USD
+  // days are treated as missing by findMissingDays and heal once the service recovers.
   const priceUnavailable = priceSource === 'unavailable'
 
   // The current day is skipped outright: fanout re-extracts it every cycle anyway, and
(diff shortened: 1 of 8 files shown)


matheus1lva:
Related yearn-prices change (open UTC day served from previous closed EOD, no DeFiLlama stampede): https://github.com/yearn/yearn-prices/pull/51

Kong #466 is the client: retry/timeout 5xx, don't fan a failed batch out to exact. yearn-prices #51 is the service: don't call Llama for the open day when yesterday's EOD is already in `token_prices`.

matheus1lva:
yearn-prices #51 no longer serves the previous closed EOD for the open UTC day. Related text above was describing that old approach.

Updated behavior: lookup requested ts → UTC EOD in `token_prices`; miss → Llama; persist today's EOD so later same-day calls are table hits (429 stampede stop). Kong-side retry/no-exact-fanout on 5xx in this PR still complements that.

PR #467, opened 2026-09-02
perf(web): cut gql vaults egress with sql merge, pagination and cache hints

matheus1lva:
### Summary
`vaults` returned every vault on every chain, filtered in JS, and shipped three overlapping jsonb
blobs per row from Postgres. Filters now run in SQL, the blobs merge in the query, keys no typeDef
serves are stripped, paging is offset based, and the hot queries carry response-cache hints.

- `vaults(chainId: Int)`: chainId stays optional (omit to list every chain). `limit: Int = 100`
  (clamped to `1..1000`) and `offset: Int = 0` added. Ordered by tvl desc, chain id asc, address asc
  (case-insensitive). Vaults with no tvl sort last as `0`. Return type unchanged.
- `buildVaultFilters` (new `lib/vaultFilters.ts`) builds the WHERE clause; the JS post-filters are gone.
  Filters read the same merged jsonb the SELECT serves. Boolean and numeric filters compare jsonb
  values directly (`= to_jsonb(...)`, `BETWEEN to_jsonb(1) AND ...`) instead of casting text with
  `::boolean` / `::numeric`, so an odd stored value filters the row out rather than failing the query.
- `apiVersion` compares the whole version as a padded `numeric[]`, not just the major. The old pattern
  opened with a non-greedy quantifier, which makes the whole regex non-greedy in postgres: `3.0.4`
  captured as `3`.
- `mergeSnapshotSql` merges `defaults || hook || snapshot` in SQL (hook `asset` wins, as before) and
  strips `VAULT_UNSERVED_KEYS`. Used by `vault`, `vaults`, `accountVaults`.
- `@cacheControl(maxAge: 300)` on `vault`, `vaults`, `timeseries`. `vault` was 3000.
- `EvmAddressSchema` gains a regex refine before its `getAddress` transform. zod does not catch throws in
  transforms, so `safeParse` was never safe and a 0x-prefixed non-address escaped as a raw viem
  `InvalidAddressError`. Also affects `ingest/extract/evmlogs.ts`, the other `safeParse` caller.
- `scripts/quality-assurance/tvl-backfill.ts` fetches its vaults with the `addresses` filter instead
  of the full list.
- `docs/graphql.md` updated for optional `chainId`, `limit`, `offset`.

Measured on the read replica for label=vault (2203 rows): blobs 11.7 MB raw, SQL merge -4%, key strip
-12% (`composition` alone is 1.27 MB).

Keyset paging (`after` cursor) was tried first and dropped: the sort key is a computed, unindexed tvl,
so keyset costs the same full sort as OFFSET at this row count while adding a cursor CTE with modes
that cannot be fixed (cursor row deleted, tvl churn between pages, ambiguous address across chains).

### How to review
- Start with `lib/vaultFilters.ts` against the deleted JS filters in
  `resolvers/vaults.ts`; each branch maps 1:1.
- `lib/mergeSnapshot.ts`: the SQL expression mirrors `mergeSnapshot()` (kept for other callers).
- `resolvers/vaults.ts`: ORDER BY + LIMIT/OFFSET. Null `chainId` must not empty the page.
- Ignore `route.ts` beyond the landing-page default query and import order.
- Smoke: `vaults { address }` then `vaults(chainId: 1) { address }` then
  `vaults(chainId: 1, offset: 100)`; pages must not overlap.

### Test plan
- [x] Manual: paged chain 1 (1302 vaults) on the Neon read replica, no overlap, tvl order held
- [x] Manual: `vaults { chainId address }` without chainId returns mixed chains
- [x] Automated: `bunx vitest run lib/mergeSnapshot.spec.ts lib/vaultFilters.spec.ts app/api/gql/resolvers/vaults.spec.ts` (24 passed)

### Risk / impact
- Breaking for clients expecting the full list (now 100 per page). Known callers: cms
  sync-vaults, tvl-backfill; tvl-backfill is updated here, cms sync-vaults must loop `offset` or pass
  a larger `limit`.
- Offset paging over a sort key the indexer keeps rewriting can duplicate or skip a row between pages.
  The 300s response cache pins pages within a walk when filters stay identical.
- `apiVersion` filtering gets stricter: a request for `>= 3.0.4` no longer matches on major alone.
- A stored JSON `null` boolean (`erc4626`, `v3`, `yearn`) reads as `false`, same as before.
- `VAULT_UNSERVED_KEYS` is a hand-picked list from the current `type Vault`; adding a typeDef field
  for one of those keys requires removing it from the list.
- `EvmAddressSchema.parse` now throws `ZodError` instead of viem's `InvalidAddressError` on a
  0x-prefixed non-address. Callers catching the viem error type specifically would miss it.
- Cache plugin only active with `GQL_ENABLE_CACHE=true`; prod env not verified from this branch.
- Rollback: revert the commits; no migrations.

### DIFF
--- a/docs/graphql.md
+++ b/docs/graphql.md
@@ -27,11 +27,11 @@ curl -s -X POST https://kong.yearn.fi/api/gql \
 
 #### `vaults`
 
-List all vaults, sorted by TVL descending.
+List a chain's vaults, one page at a time, sorted by TVL descending then address ascending.
 
 | Argument | Type | Description |
 |----------|------|-------------|
-| `chainId` | `Int` | Filter by chain |
+| `chainId` | `Int!` | Chain to list (required) |
 | `addresses` | `[String]` | Filter by addresses |
 | `apiVersion` | `String` | Minimum API version (e.g. `"3.0.0"`) |
 | `erc4626` | `Boolean` | ERC-4626 vaults only |
@@ -41,12 +41,19 @@ List all vaults, sorted by TVL descending.
 | `vaultType` | `Int` | Filter by vault type |
 | `riskLevel` | `Int` | Filter by risk level |
 | `unratedOnly` | `Boolean` | Only unrated vaults |
+| `limit` | `Int` | Page size, default `100`, clamped to `1..1000` |
+| `after` | `String` | Address of the last vault of the previous page |
+
+Paging is keyset, not offset. Pass the `address` of the last row you received as
+`after` to get the next page, repeating with the same filters until a page comes
+back empty. Vaults with no TVL sort last as `0`. An `after` that matches no vault
+on the chain returns an empty page, same as the end of the list.
 
 Returns [`Vault`](#vault-1).
 
 ```graphql
 {
-  vaults(chainId: 1, v3: true) {
+  vaults(chainId: 1, v3: true, limit: 50, after: "0x1234...") {
     address
     name
     symbol
--- a/docs/graphql.md
+++ b/docs/graphql.md
@@ -27,7 +27,7 @@ curl -s -X POST https://kong.yearn.fi/api/gql \
 
 #### `vaults`
 
-List vaults one page at a time, sorted by TVL descending then address ascending. Omit `chainId` to list every chain.
+List vaults one page at a time, sorted by TVL descending, address ascending, then chain ID ascending. Omit `chainId` to list every chain.
 
 | Argument | Type | Description |
 |----------|------|-------------|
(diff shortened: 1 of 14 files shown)


murderteeth (approved):
## Summary
Replaces the JS vault filters and merge with a SQL filter builder and a merged-blob query, adds offset pagination to `vaults`, and sets cache-control hints on the vault resolvers. The SQL filter builder maps one-to-one onto the deleted JS filters, and the merged-blob SQL reproduces the JS merge exactly on a sample of live rows.

## Issues
None.

## Verdict
APPROVE

Reviewed at 0c10deae. The later commit c1d8637 is test-only and passes against that source.

---

## How This Was Reviewed

matheus1lva:
Ok now im gonna hold on merging until katana people give a thumbs up that they are ready for this to be merged - just in case.

PR #469, opened 2026-09-04
feat: add robinhood chain (4663)

matheus1lva:
### Summary
Adds Robinhood Chain (chain ID 4663) as a supported indexing chain, plus a manual entry
for its spUSDG/USDG Morpho looper vault.

### How to review
- `packages/lib/chains.ts` / `packages/web/chains.ts`: new `robinhood` chain def, RPC and
  block explorer URLs, multicall3 address.
- `packages/lib/multicall3.ts`: activation block wired for the new chain.
- `packages/ingest/prices.ts`: added to the price-service chain-name map.
- `config/manuals.yaml`: manual vault entry for `0xC7d60aBfa6f4D79C2C42Bf84A54795bD8c586957`.
- `.env.example`: new `HTTP_ARCHIVE_4663` / `HTTP_FULLNODE_4663` vars.

### Test plan
- [ ] Manual, local integration:
  1. In `.env`, set `HTTP_ARCHIVE_4663` and `HTTP_FULLNODE_4663`
     (public RPC: `https://rpc.mainnet.chain.robinhood.com`).
  2. `make dev`, wait for redis/postgres/ingest/web to come up.
  3. In the terminal pane: `ingest` → `fanout abis`. Run it twice if the first pass
     only seeds manuals.
  4. Check the vault landed:
     `select chain_id, address, label from thing where chain_id = 4663;`
     expect `0xC7d60aBfa6f4D79C2C42Bf84A54795bD8c586957` with label `vault`.
  5. Check snapshot/evmlog rows exist for chain 4663 and `evmlog_strides` covers blocks
     from `inceptBlock` 3362381 onward.
  6. Query http://localhost:3001/api/gql for the vault on chainId 4663 and confirm
     name `spUSDG/USDG Morpho Looper`, asset `0x5fc5…d168`, decimals 6.
  7. Prices: with `USE_PRICE_SERVICE` on, confirm no "unavailable price" loop in ingest
     logs for chain 4663 (chain-name map entry `robinhood`).
- [ ] Automated (when applicable): not run.

### Risk / impact
Additive only, no changes to existing chains. Rollback: remove `robinhood` from
`config/chains.yaml`.


murderteeth (dismissed):
The chain definition itself checks out against the live chain: chain id, multicall3 predeploy, explorer, vault creation block, incept timestamp, asset, decimals and name all match.

### Issues

- **Manual entry contradicts the on-chain contract (high)** — the vault is a Yearn TokenizedStrategy (apiVersion 3.0.4, share symbol `ysUSDG`) but is recorded as non-Yearn, non-v3 with no `apiVersion`, so none of the yearn/3 hooks (reports, fees, performance) run for it and it is absent from every v3- or yearn-filtered API query, which is how most consumers find vaults. `config/manuals.yaml:213`
  Done when: the thing's defaults match the contract, and `vaults(chainId: 4663, v3: true)` and `vaults(chainId: 4663, yearn: true)` both return it.
  Provenance: a7e2d62

- **Price service does not know the new chain name (high, handled in yearn-prices)** — every price lookup tested for chain 4663 fails with a server error, in both the exact and batch routes, the same response a made-up chain name gets. Until that lands the vault gets no USD TVL, and because the batch queue is shared across chains, one Robinhood coin in a flush window fails the whole `batchHistorical` request and pushes every other chain's coins in that window to per-coin fallback calls. The fix is being tracked in the yearn-prices project, not here; this PR just needs to keep `4663` mapped to the name that service ends up using. `packages/ingest/prices.ts:208`
  Done when: the historical price route for `robinhood:0x5fc5…d168` returns 200 or 404 rather than 500, and a mixed batch containing a Robinhood coin returns 200. Test plan step 7 should be re-run once that is true.
  Provenance: a7e2d62

---

## How This Was Reviewed

murderteeth (commented):
The manual entries now match the contract (TokenizedStrategy 3.0.4, `ysUSDG`, USDG asset, 6 decimals) and prices.yearn.dev now resolves the `robinhood` chain name, so both findings from the previous round are closed.

### Issues

- **APR oracle never indexed on Robinhood (medium)** — the Yearn APR oracle is deployed on chain 4663 at the canonical address and answers for this vault, but kong's per-chain oracle table has no 4663 entry, so the timeseries hook silently writes no `apr-oracle` rows and the vault's oracle APR/APY stays at the zero fallback, with no log or error to flag it. This is not new since the last round: it was already reachable at a7e2d62 via the erc4626 path and our first review missed it. Every other chain in `config/chains.yaml` has an entry, and the Bera/Katana onboarding needed exactly this follow-up fix (#315). `packages/ingest/abis/yearn/3/vault/timeseries/apr-oracle/constants.ts:46`
  Done when: after a timeseries fanout for chain 4663, `select * from output where chain_id = 4663 and label = 'apr-oracle'` returns rows for `0xC7d60aBfa6f4D79C2C42Bf84A54795bD8c586957`, and the entry's `inceptBlock` is not earlier than the oracle's deployment block on Robinhood.
  Provenance: pre-existing 08ee3a3 (PR depends on it via `config/chains.yaml` a7e2d62 and the manual entry)

---

## How This Was Reviewed

murderteeth (approved):
The apr-oracle entry added in 3b119c3 checks out: block 3354374 is the oracle's deployment transaction on Robinhood, and both round-2 findings are closed. No issues.

---

## How This Was Reviewed

PR #472, opened 2026-09-08
Read current vault allocators from Role Managers

### LINKED ISSUE #471: Incorrect V3 allocator discovery and GraphQL resolution after Role Manager updates
## Summary

Kong’s GraphQL API can return a stale debt allocator after a V3 Role Manager changes the allocator assigned to a vault.

Kong currently derives the allocator from the vault-bound factory event:

```solidity
NewDebtAllocator(address allocator, address vault)
```

It does not use `AddedNewVault` or `UpdateDebtAllocator` as active-assignment evidence. Kong also does not currently configure the shared debt allocator factory or shared allocator ABI.

## Confirmed example

Ethereum yvUSDC-1:

```text
Vault:
0xBe53A109B494E5c9f97b9Cd39Fe969BE68BF6204

Role Manager:
0xb3bd6B2E61753C311EFbCF0111f75D29706D9a41

Original vault-bound allocator:
0x1400D5C76D0A630368c8548172e5130983AAA0Ef

Allocator assigned by UpdateDebtAllocator:
0x1e9eB053228B1156831759401dE0E115356b8671
```

The replacement happened in block `20,987,762` through:

```solidity
UpdateDebtAllocator(
    vault: 0xBe53A109B494E5c9f97b9Cd39Fe969BE68BF6204,
    debtAllocator: 0x1e9eB053228B1156831759401dE0E115356b8671
)
```

### GraphQL vault lookup

```graphql
query VaultAllocator($chainId: Int!, $address: String!) {
  vault(chainId: $chainId, address: $address) {
    address
    allocator
    role_manager
  }
}
```

Variables:

```json
{
  "chainId": 1,
  "address": "0xBe53A109B494E5c9f97b9Cd39Fe969BE68BF6204"
}
```

Current response:

```json
{
  "data": {
    "vault": {
      "address": "0xBe53A109B494E5c9f97b9Cd39Fe969BE68BF6204",
      "allocator": "0x1400D5C76D0A630368c8548172e5130983AAA0Ef",
      "role_manager": "0xb3bd6B2E61753C311EFbCF0111f75D29706D9a41"
    }
  }
}
```

The returned allocator is the original allocator, not the allocator subsequently assigned by the Role Manager.

## Top-level `allocator` query issue

The dedicated GraphQL field also does not currently return the allocator contract address.

```graphql
query Allocator($chainId: Int!, $vault: String!) {
  allocator(chainId: $chainId, vault: $vault) {
    chainId
    address
    vault
  }
}
```

Variables:

```json
{
  "chainId": 1,
  "vault": "0xBe53A109B494E5c9f97b9Cd39Fe969BE68BF6204"
}
```

Current response:

```json
{
  "data": {
    "allocator": {
      "chainId": 1,
      "address": "0xfCF8c7C43dedd567083B422d6770F23B78D15BDe",
      "vault": "0xBe53A109B494E5c9f97b9Cd39Fe969BE68BF6204"
    }
  }
}
```

Here, `address` is the factory contract that emitted `NewDebtAllocator`, not the allocator address contained in the event arguments.

The resolver currently selects `evmlog.address` instead of `args->>'allocator'`.

## Current indexing behavior

Allocator discovery currently works as follows:

1. Kong indexes configured vault-bound `DebtAllocatorFactory` contracts.
2. A `NewDebtAllocator(allocator,vault)` event causes the factory hook to create a `debtAllocator` thing.
3. Kong indexes that address using its vault-bound debt allocator ABI.
4. `projectDebtAllocator()` resolves a vault’s allocator by finding the latest factory event whose `vault` argument matches the requested vault.

Relevant files:

- `config/abis.yaml`
- `packages/ingest/abis/yearn/3/debtManagerFactory/event/hook.ts`
- `packages/ingest/abis/yearn/3/roleManager/event/hook.ts`
- `packages/ingest/abis/yearn/3/vault/snapshot/hook.ts`
- `packages/web/app/api/gql/resolvers/allocator.ts`

The Role Manager ABI includes `UpdateDebtAllocator`, but its event hook currently subscribes only to `AddedNewVault`. The `debtAllocator` argument from `AddedNewVault` is parsed but not used for allocator discovery.

## Root cause

Factory creation is being treated as the source of the active vault-to-allocator association.

These are separate concepts:

- `NewDebtAllocator` proves how and when an allocator was deployed.
- `AddedNewVault` establishes the initial active assignment.
- `UpdateDebtAllocator` changes the active assignment.

The current projection therefore misses:

- Allocators assigned through `UpdateDebtAllocator`.
- Shared allocators whose factory event contains governance rather than a vault.
- Manually deployed allocators without a recognized factory event.
- Allocators created by future or unconfigured factory variants.

## Expected behavior

The GraphQL `vault.allocator` field should return the allocator currently assigned to that vault.

The top-level `allocator(chainId,vault)` query should return the allocator contract address, not the factory address.

For historical projections, Kong should resolve assignments in deterministic event order:

1. `AddedNewVault(vault,debtAllocator,category)` establishes the initial assignment.
2. `UpdateDebtAllocator(vault,debtAllocator)` replaces it.
3. Factory events remain deployment and ABI-provenance evidence.

Shared allocators should be indexed with their correct ratio event:

```solidity
UpdateStrategyDebtRatio(
    address indexed vault,
    address indexed strategy,
    uint256 newTargetRatio,
    uint256 newMaxRatio,
    uint256 newTotalDebtRatio
)
```

## Suggested changes

**Note that these changes and the fix to this issue may overlap with the changes to be made to index and present allocator data that is ongoing.**

- Process `UpdateDebtAllocator` in the Role Manager event hook.
- Use both `AddedNewVault` and `UpdateDebtAllocator` to discover assigned allocator addresses.
- Enqueue newly assigned allocator contracts for indexing even when no recognized factory event exists.
- Configure the shared debt allocator factory and shared allocator ABI.
- Resolve `vault.allocator` from the latest ordered assignment.
- Fix the top-level `allocator` resolver so `address` is the allocator address.
- Retain factory events as deployment provenance.
- If an assigned allocator’s ABI cannot be identified, preserve the assignment and expose it as unsupported rather than retaining the previous allocator silently.

## GraphQL behavior after the fix

This query:

```graphql
query VaultAllocator {
  vault(
    chainId: 1
    address: "0xBe53A109B494E5c9f97b9Cd39Fe969BE68BF6204"
  ) {
    address
    allocator
  }
}
```

Should return:

```json
{
  "data": {
    "vault": {
      "address": "0xBe53A109B494E5c9f97b9Cd39Fe969BE68BF6204",
      "allocator": "0x1e9eB053228B1156831759401dE0E115356b8671"
    }
  }
}
```

The dedicated query should agree:

```graphql
query Allocator {
  allocator(
    chainId: 1
    vault: "0xBe53A109B494E5c9f97b9Cd39Fe969BE68BF6204"
  ) {
    chainId
    address
    vault
  }
}
```

Expected allocator address:

```text
0x1e9eB053228B1156831759401dE0E115356b8671
```

## Acceptance criteria

- `vault.allocator` returns the latest allocator assigned through the Role Manager.
- `allocator(chainId,vault).address` returns the allocator contract, not its factory.
- yvUSDC-1 returns `0x1e9eB053228B1156831759401dE0E115356b8671`.
- `UpdateDebtAllocator` causes subsequent GraphQL queries to return the replacement.
- An assigned allocator without known factory provenance is discovered or explicitly marked unsupported.
- Shared allocator contracts and ratio events are indexed.
- Historical resolution respects block number, transaction index, and log index.
- Factory deployment records remain available as provenance.
- Tests cover initial assignments, replacements, shared allocators, unknown provenance, and multiple vaults sharing one allocator.

rossgalloway:
### Summary

Kong can report an old vault allocator after replacement, and its dedicated GraphQL query can return the factory address instead. Read the assignment from the vault's Role Manager contract during snapshots, then serve the saved allocator and strategy ratios consistently across APIs.

Addresses the current-state problems in #471 and removes the untrusted event lookup described in #476. The scope separates two capabilities:

- **Current state — covered by this PR:** Ask the vault's Role Manager which allocator is assigned, read its strategy ratios, and save the result in the usual snapshot. This supports shared allocators too.
- **Historical tracking — separate work:** Index shared allocators' ratio-change events and reconstruct which allocator was assigned at each point in the past. This is not required for the current-state fix.

Issue #471 also includes those historical tracking requirements. This PR does not automatically close it; the remaining requirements need to be tracked separately or explicitly removed from its scope first.

### How to review

1. Start with `packages/ingest/abis/yearn/lib/allocator.ts` and the V3 vault snapshot hook. Assignment and ratio reads use the snapshot block. Both supported ratio interfaces are tried; complete valid pairs are accepted, valid zeros are preserved, and conflicting or unsupported results become null. Transport failures retry without saving partial data or using an event fallback.
2. Review `packages/ingest/load/index.ts`. Older V3 snapshots cannot overwrite newer ones. Three documented legacy Gnosis vault/controller pairs preserve saved allocator fields while refreshing accounting. A controller change ends that exception; missing historical values are not reconstructed.
3. Check `packages/web/app/api/gql/resolvers/allocator.ts`. The dedicated query now reads the same saved snapshot as the vault and REST APIs.
4. Review `allocator-security.spec.ts` in the V3 vault snapshot directory. It proves that a newer forged deployment event would fool the old query but cannot change the new snapshot or its RPC targets.

See `docs/allocator-assignments.md` for the rollout steps and `docs/allocator-476-audit.md` for the Gnosis audit and its limits. Indexing stays in Kong. The implementation uses existing snapshot fields, without Envio, a separate allocator state model, or assignment-event reconstruction.

### Test plan

- [x] Automated: 61 ingest mock tests, 14 targeted PostgreSQL-backed tests, and 90 web tests passed. Coverage includes interface selection, replacements and clears, transport failures, legacy preservation, concurrent writes, API agreement, and forged events.
- [x] Automated: Web production build and package lint checks passed. Ingest type-check output matches the unchanged base's 17 diagnostics.
- [x] Manual: Read-only Ethereum checks at blocks 20,987,761 and 20,987,762 reproduced the allocator replacement and corresponding ratios. Ethereum and Polygon checks covered valid zero and nonzero ratios; Polygon's current checks explicitly used its historical strategy list because its current default queue was empty.
- [x] Manual: On 2026-09-18, inspect the three legacy Gnosis vaults through public REST/GraphQL APIs and verify controllers at the served snapshot block. No non-null allocator addresses or ratios were exposed.
- [ ] Deployment: Inventory the raw Gnosis database rows and verify provenance for any non-null legacy values. Deploy ingestion, refresh affected snapshots, compare Gnosis fields, then promote the API change and refresh caches. Verify agreement across REST and both GraphQL paths.

### Risk / impact

This changes reported allocator addresses and strategy ratios. It does not submit transactions or move funds, and it requires no database migration. Existing incorrect snapshots remain visible until refreshed; API caches can delay the corrected values.

Raw production database access was unavailable during the Gnosis audit. The public API observations do not certify historical data or raw stored values. Complete the database check before preserving non-null legacy values in production.

No production rollout has been performed. If validation fails during rollout, pause affected snapshot jobs and withhold API promotion. Existing fields remain compatible with the prior deployment, but reverting ingestion would restore the old lookup and its known defects; retain the before/after inventory for any recovery.


matheus1lva (commented):
Fixes #471's core regression correctly; the ordered resolver and shared-projection design are sound.

- **Allocator regression on Gnosis and Polygon (high)** — V3 vaults on chains 100 and 137 lose `vault.allocator` and all target/max ratios: the factory-derived path is deleted, and the new projection returns `chain_not_configured` for anything outside the hardcoded chain list.
  Done when: a Polygon or Gnosis V3 vault with a `NewDebtAllocator` log still returns its allocator after the snapshot hook runs, or the PR states those chains are intentionally dropped.
  Provenance: 12c83eb

- **Transient outage blanks served data (high)** — any Envio or RPC failure during the routine snapshot hook replaces a previously validated projection with an unavailable one, so the vault's allocator and every strategy ratio read null in GraphQL and REST until the next successful run. This is documented as intended, but it is an availability regression versus main, which served the address regardless of Envio.
  Done when: with a validated projection stored, a hook run whose Envio call throws leaves the served allocator and ratios unchanged, at most with a stale marker.
  Provenance: 12c83eb

- **Unmet acceptance criterion (medium)** — #471 requires shared allocator contracts and their `UpdateStrategyDebtRatio` events to be indexed in Kong; this PR delegates discovery to Envio and reads ratios as point-in-time RPC calls, without amending the issue scope.
  Done when: Kong indexes shared allocator ratio events, or the issue is explicitly amended to move that out of scope.
  Provenance: 12c83eb

- **Missing shared-allocator multi-vault test (medium)** — #471 lists "multiple vaults sharing one allocator" as a required test case; every spec uses a single vault constant.
  Done when: a test resolves two vaults against the same shared allocator and asserts each projection's address and ratio calls are scoped to its own vault.
  Provenance: 12c83eb

- **Pagination untested (low)** — the keyset cursor, stall detection, and event cap in the Envio query have no coverage; every fixture returns one page under the page size.
  Done when: a test feeds more than one page and asserts the resolved allocator comes from the last page, plus one test each for the stalled-cursor and event-limit throws.
  Provenance: 12c83eb

- **Deployment rejection branches untested (low)** — wrong factory address, vault-bound row on Katana, and ambiguous deployments are all swallowed into `allocator_configuration_unavailable`, so a regression there is invisible.
  Done when: tests cover each rejection branch and one valid Katana shared deployment.
  Provenance: 12c83eb

---

## How This Was Reviewed

murderteeth (changes requested):
## Summary
An independent multi-lens pass surfaced the same defects matheus1lva already posted (Polygon regression, outage blanking, unmet #471 criteria, missing shared-allocator and pagination tests), so those are not repeated. Two findings below: the first sharpens matheus1lva's "Unmet acceptance criterion" with replica evidence for why Envio is unnecessary, the second is new.

## Issues
- [ ] **packages/ingest/allocators.ts:44** - **Envio is not needed; this is a Kong config gap (high)** — same defect as "Unmet acceptance criterion (medium)" in matheus1lva's review, raised to high because every input the projection fetches from Envio is data Kong already knows how to index. The roleManager ABI carries `AddedNewVault`, `RemovedVault` and `UpdateDebtAllocator`, and the extractor stores every ABI event, not just hooked ones; the vault ABI carries `UpdateRoleManager`; vault-bound factory deployments are already in `evmlog`. What Kong is missing on prod is coverage, not capability: the yvUSDC-1 role manager (`0xb3bd…`) predates the roleManagerFactory, so it is not a `thing` and Kong holds zero logs from it, which is the entire #471 case; and the shared factory (`0x03D4…`) is not in `abis.yaml`. Both close with config: legacy role managers declared in `config/manuals.yaml`, the shared factory added as a source with its allocator ABI. The ordered resolver this PR adds in `lib/allocators.ts` can then run over Kong's own `evmlog`. Instead the PR introduces a second indexer, three env vars, a 30-second external call per vault snapshot, and a projection that goes null whenever that indexer is unreachable.
  - Done when: allocator assignments are resolved from Kong's own `evmlog` with no `ENVIO_*` configuration; the legacy role managers and the shared factory are indexed via config; yvUSDC-1 resolves to `0x1e9eB053…` from Kong data alone; and `UpdateDebtAllocator` rows exist in `evmlog` for chain 1.
  - Provenance: 12c83eb
- [ ] **packages/ingest/allocators.ts:1** - **Feature code placed at the package root (medium)** — the ingest root holds only cross-cutting infrastructure (db, rpcs, prices, things, the service entry), and protocol logic lives under `abis/`, with `abis/yearn/lib/` already serving as the shared home for Yearn modules. This PR puts three Yearn-specific files at the root instead: the projection, whose only consumer is the V3 vault snapshot hook reaching up five directory levels; the store helper, used only by the CLI; and a one-off rollout CLI, for which the repo already has `packages/scripts/src/`. Anything at the root reads as shared infrastructure to `packages/scripts` importers, so this placement will be copied.
  - Done when: no new `.ts` files sit at the ingest package root; the projection lives under `abis/yearn/`, the CLI and its store helper live under `packages/scripts/src/` or beside the feature they serve, the specs follow their modules, and the docs' invocation path is updated to match.
  - Provenance: 12c83eb

## Verdict
REQUEST_CHANGES

---

## How This Was Reviewed

rossgalloway:
fixes have been applied and the description has been updated.

matheus1lva:
Verified these three against the branch. All reproduce in the code as written.

### 1. Manager change keeps serving the old allocator (high)

`projectCurrentAllocator` returns early at `packages/ingest/abis/yearn/lib/allocators/projection.ts:126` for every
`unavailable` result except `vault_removed_from_role_manager` — before `revision` is computed at
`projection.ts:164`. So the projection carries `revision: null`.

`mergeAllocatorHook` classifies `revision == null` as a failed refresh
(`packages/lib/allocator-snapshot.ts:51`) and retains the previous state with `stale: true`
(`allocator-snapshot.ts:56`). When a vault's role manager moves to an address with no indexed assignment
events (EOA, third-party manager, zero), every subsequent refresh repeats this, so the previous manager's
allocator is served indefinitely through both GraphQL and REST.

The projection-level behaviour is covered at `projection.mock.spec.ts:121`, but nothing covers the merge
retention for the manager-change case.

Conflating "manager not indexed" with "refresh failed" is defensible — Kong cannot see logs of an unconfigured
manager. But the happy path already calls `getDebtAllocator(vault)`; calling it on this path disambiguates:
a zero return from the new manager is conclusive, not a failure.

Done when: with a validated `allocatorState` stored, a manager change to an address with no assignment rows
yields `allocator: null` or an explicit unavailable state, transient RPC/DB failures still retain, and a test
covers the manager-change merge.

### 2. Every unmaterialized vault answers null on deploy (medium)

`allocatorSnapshotFields` forces `allocator: null` when `allocatorState` is absent
(`packages/lib/allocator-snapshot.ts:17`), and its SQL twin does the same. `mergeSnapshot` applies it after the
hook spread, so from the web deploy every V3 vault serves no allocator until its snapshot hook re-runs.

The removed `projectDebtAllocator` read the latest `NewDebtAllocator` by vault — a deployment, not an
assignment — which is exactly what this PR is fixing. Still, the interim needs to be a decision, not a
side effect. Combined with (1), vaults whose role manager is never indexed stay null permanently.

Done when: a snapshot with a legacy `hook.allocator` and no `allocatorState` still serves that address, or the
rollout sequences backfill before the web deploy and states the interim null as accepted risk.

### 3. Raw DB and viem errors reach GraphQL clients (low)

`packages/web/app/api/gql/resolvers/allocator.ts:7-13` drops the try/catch that the sibling `vault` resolver
keeps, and the server at `packages/web/app/api/gql/route.ts:38` sets no `formatError`. A malformed `vault`
argument throws viem's `InvalidAddressError` and a pg failure serializes its message to unauthenticated
callers. Low impact — the leaked content is an address or a pg message — but the inconsistency with `vault` is
free to fix.

Done when: a DB failure or malformed address on `allocator(chainId, vault)` returns the same generic error as
`vault`, with details only in server logs.

Priority: fix (1), sequence or annotate (2), (3) is a two-line wrap.

rossgalloway:
> ### 2. Every unmaterialized vault answers null on deploy (medium)
> 
> `allocatorSnapshotFields` forces `allocator: null` when `allocatorState` is absent (`packages/lib/allocator-snapshot.ts:17`), and its SQL twin does the same. `mergeSnapshot` applies it after the hook spread, so from the web deploy every V3 vault serves no allocator until its snapshot hook re-runs.
> 
> The removed `projectDebtAllocator` read the latest `NewDebtAllocator` by vault — a deployment, not an assignment — which is exactly what this PR is fixing. Still, the interim needs to be a decision, not a side effect. Combined with (1), vaults whose role manager is never indexed stay null permanently.
> 
> Done when: a snapshot with a legacy `hook.allocator` and no `allocatorState` still serves that address, or the rollout sequences backfill before the web deploy and states the interim null as accepted risk.

Codex has a question:
On the deployment/backfill issue, I'd like to clarify how you'd prefer this handled before I make a change.

My concern with adding a legacy fallback is that the old factory-derived allocator isn't simply an older representation of the same data — it can be incorrect, which is one of the things this PR is intended to fix. I'd prefer not to knowingly preserve two conflicting sources of truth.

Would you prefer that we handle this as a staged migration: deploy the new ingestion/indexing path first, backfill and verify `allocatorState` coverage, and only then switch the API/web readers to the new projection?

If ingestion and web can't practically be deployed/cut over separately, would you instead prefer an explicit activation/feature gate so the new reader isn't enabled until the backfill has been verified?

Or do you specifically want the legacy factory-derived allocator retained as a temporary fallback during the migration?

rossgalloway:
I'll push fixes for the other 2 issues shortly

matheus1lva:
> > ### 2. Every unmaterialized vault answers null on deploy (medium)
> > `allocatorSnapshotFields` forces `allocator: null` when `allocatorState` is absent (`packages/lib/allocator-snapshot.ts:17`), and its SQL twin does the same. `mergeSnapshot` applies it after the hook spread, so from the web deploy every V3 vault serves no allocator until its snapshot hook re-runs.
> > The removed `projectDebtAllocator` read the latest `NewDebtAllocator` by vault — a deployment, not an assignment — which is exactly what this PR is fixing. Still, the interim needs to be a decision, not a side effect. Combined with (1), vaults whose role manager is never indexed stay null permanently.
> > Done when: a snapshot with a legacy `hook.allocator` and no `allocatorState` still serves that address, or the rollout sequences backfill before the web deploy and states the interim null as accepted risk.
> 
> Codex has a question: On the deployment/backfill issue, I'd like to clarify how you'd prefer this handled before I make a change.
> 
> My concern with adding a legacy fallback is that the old factory-derived allocator isn't simply an older representation of the same data — it can be incorrect, which is one of the things this PR is intended to fix. I'd prefer not to knowingly preserve two conflicting sources of truth.
> 
> Would you prefer that we handle this as a staged migration: deploy the new ingestion/indexing path first, backfill and verify `allocatorState` coverage, and only then switch the API/web readers to the new projection?
> 
> If ingestion and web can't practically be deployed/cut over separately, would you instead prefer an explicit activation/feature gate so the new reader isn't enabled until the backfill has been verified?
> 
> Or do you specifically want the legacy factory-derived allocator retained as a temporary fallback during the migration?

For instance - lets move to packages/scripts/src/ given thats where most of the ad-hoc scritps live on:

- packages/ingest/abis/yearn/lib/allocators/refresh.ts — the CLI. Only entrypoint.
- packages/ingest/abis/yearn/lib/allocators/refresh.mock.spec.ts — its test.
- packages/ingest/abis/yearn/lib/allocators/store.ts + store.spec.ts

Stays in ingest (runtime feature code, imported by the vault snapshot hook):

- allocators/projection.ts + projection.mock.spec.ts
- allocator-discovery.ts + spec, allocator-indexing.spec.ts
- lib/allocator-snapshot.ts, lib/allocators.ts, lib/allocator-types.ts (shared with web)

Also update the docs invocation path in docs/allocator-assignments.md step 4.

Re review: thats fair to me - i think the only piece remaining is the "post merge" steps in the pr description for eg, generated by claude:

```
 deploy ingest, run discovery/fanout for the
manual role managers + shared factory, complete the evmlog backfill, run the refresh CLI
with --write, then promote the web build. Note that until backfill completes a re-run hook
still writes an unavailable state, so the window is "backfill done", not "ingest deployed".
State the interim null as accepted risk with that window.
``` 

So when i do the deployment i know what is the correct steps to run!

matheus1lva (approved):

rossgalloway:
A more succinct description of what this PR does and why:

## The problem being solved

A debt allocator holds target and maximum allocation ratios for a vault’s strategies. Those settings help consumers understand the intended distribution of capital; they are separate from the actual amounts currently allocated.

**Kong previously looked for a factory event saying an allocator had been created for a vault. That tells us about deployment, but it does not tell us whether the vault’s Role Manager subsequently assigned a replacement. Shared allocators also break the old assumption: one allocator can serve several vaults, and its factory event identifies governance rather than a particular vault.**

The reported Ethereum yvUSDC-1 example illustrates both problems. Its manager replaced allocator `0x1400…A0Ef` with shared allocator `0x1e9e…8671` at block 20,987,762. The vault API still reported the original address. Separately, the dedicated allocator query selected the event emitter’s address, returning the factory itself. These are the reproductions recorded in [issue #471](https://github.com/yearn/kong/issues/471), rather than a fresh production observation.

## What changes architecturally

| Concern | Before | After |
| --- | --- | --- |
| Assignment authority | Infer the allocator from factory creation. | Resolve assignments from the current Role Manager’s events and verify against contract state. |
| Allocator model | Assume a deployment belongs to one vault. | Track each vault’s assignment independently, including several vaults sharing one allocator. |
| API data flow | Separate factory lookup and snapshot logic. | GraphQL and REST derive allocator fields from one saved per-vault state. |
| Failure semantics | Legacy fields cannot express the new evidence and freshness distinctions. | Explicit assignment status, interface support, observation block, revision, and stale metadata. |

murderteeth (commented):
Can we simplify this PR around correcting the current allocator and ratios like this:

1. Read `getDebtAllocator(vault)` from the vault’s current role manager. For legacy managers that lack this getter, retain the existing factory-event lookup as a fallback. A transient RPC failure should not trigger that fallback.
2. Multicall both ratio getter signatures—`(strategy)` and `(vault, strategy)`—with `allowFailure: true`. Use the successful interface’s target/max pair, preserving valid zeros.
3. Save the existing snapshot fields.
4. Change the top-level GraphQL `allocator` resolver to read that snapshot. `vault.allocator` and REST already consume it.

We verified direct contract reads against yvUSDC-1 before and after its allocator replacement, and successfully read the replacement’s current ratios.

Trying both interfaces removes the need to classify allocators as shared or vault-bound, including the factory-provenance lookup used for that classification. Failed or conflicting results still need explicit handling.

For this fix, I don’t think we need the new `allocatorState` model, historical-event prerequisite, or custom refresh/migration machinery.

rossgalloway:
Seems like these are good changes. But Codex and I have a few questions:

> 1. Read `getDebtAllocator(vault)` from the vault’s current role manager. For legacy managers that lack this getter, retain the existing factory-event lookup as a fallback. A transient RPC failure should not trigger that fallback.

[RG]: Do we want to keep the fallback? If we do keep the fallback, there seems to be a chance that we could get incorrect data about the current allocator, if the allocator has changed since the factory-event. It seems like the only vaults where we need the fallback are the Gnosis Chain vaults, which are all retired at this point.

> 3. Save the existing snapshot fields.

[codex]:
For this scope, I agree we can probably remove the event reconstruction, factory classification, allocatorState model and dedicated materialization command. However, with their removal, we should be deliberate about the guarantees being dropped. The current implementation contains protections for outages, confirmed clears, replacements and delayed writes. Those concerns still exist with direct reads. Do we still want a  smaller implementation still needs to ensure that:

- Confirmed clears actually overwrite the previous allocator.
- A replacement never receives its predecessor’s ratios.
- Failed reads are not silently converted into valid zeros.
- Older snapshot jobs cannot undo a newer observation.

Should we use existing snapshot behavior where it is sufficient and add only narrowly necessary protection? Removing the custom stale metadata would also mean accepting ordinary snapshot freshness semantics instead of the richer guarantees described in our document.

murderteeth:
> [RG]: Do we want to keep the fallback? If we do keep the fallback, there seems to be a chance that we could get incorrect data about the current allocator, if the allocator has changed since the factory-event. It seems like the only vaults where we need the fallback are the Gnosis Chain vaults, which are all retired at this point.

i'm fine with no longer supporting the legacy fallback as long as the remaining gnosis vault allocator snapshots are preserved and not overwritten.

> [codex]... Should we use existing snapshot behavior where it is sufficient and add only narrowly necessary protection?

yes

rossgalloway:
This PR has been updated to take into account the requested changes. The Description has also been updated.

packages/ingest/abis/yearn/3/vault/snapshot/hook.ts
```diff
@@ -97,13 +99,29 @@ type Snapshot = z.infer<typeof SnapshotSchema>
...
 export default async function process(chainId: number, address: `0x${string}`, data: any) {
   const snapshot = SnapshotSchema.parse(data)
-  const strategies = await projectStrategies(chainId, address, undefined, snapshot)
+  let strategies = await projectStrategies(chainId, address, snapshot.blockNumber, snapshot)
+  const legacyAllocator = isLegacyAllocator(chainId, address, snapshot.role_manager)
+  if (legacyAllocator) {
+    // Retired strategies may no longer be in the queue/events. Re-read their
+    // accounting so preserving their stored ratios never drops their rows.
+    const { rows } = await db.query(`SELECT hook FROM snapshot WHERE chain_id = $1 AND address = $2
+      AND lower(snapshot->>'role_manager') = lower($3)`, [chainId, address, snapshot.role_manager])
```
matheus1lva, line 108:
```suggestion
      AND snapshot->>'role_manager' = $3`, [chainId, address, getAddress(snapshot.role_manager)])
```

The samples i pulled are already checksummed. Also dont forget to import getAddress from viem

murderteeth (changes requested):
The core fix is right. Reading `getDebtAllocator(vault)` from the Role Manager and pointing the top-level resolver at the snapshot fixes both #471 bugs, and removing the old event lookup closes #476. Mainnet check: the new read returns shared allocator `0x1e9e…8671` for all 12 vaults under `0xb3bd…`.

### Issues

- [ ] **packages/ingest/load/index.ts:108** — **Scope creeps beyond fix (high)** — the fix needs about 50 lines in two places, but the PR ships 14 files. The extra code changes every V3 snapshot write, not just allocator fields: older-block writes are now silently dropped for all V3 vaults. The legacy Gnosis branch preserves values that are empty: on prod, all three vaults store no allocator and no target/max ratios. Reviewing and maintaining it costs more than the fix itself.
  Done when: the diff is limited to the allocator read in the vault snapshot hook and `lib/allocator.ts`, the top-level `allocator` resolver and its typeDef, and their tests (`allocator.mock.spec.ts`, `hook.spec.ts`, `extract/snapshot.spec.ts`, resolver spec). The `load/index.ts` changes, the legacy Gnosis branch, `load/allocator.spec.ts`, `allocator-security.spec.ts` and the three `docs/allocator-*` files are removed.
  Provenance: 3609121, 89d36d9

rossgalloway:
> The core fix is right. Reading `getDebtAllocator(vault)` from the Role Manager and pointing the top-level resolver at the snapshot fixes both #471 bugs, and removing the old event lookup closes #476. Mainnet check: the new read returns shared allocator `0x1e9e…8671` for all 12 vaults under `0xb3bd…`.
> ### Issues
> 
>     * [ ]  **packages/ingest/load/index.ts:108** — **Scope creeps beyond fix (high)** — the fix needs about 50 lines in two places, but the PR ships 14 files. The extra code changes every V3 snapshot write, not just allocator fields: older-block writes are now silently dropped for all V3 vaults. The legacy Gnosis branch preserves values that are empty: on prod, all three vaults store no allocator and no target/max ratios. Reviewing and maintaining it costs more than the fix itself.
>       Done when: the diff is limited to the allocator read in the vault snapshot hook and `lib/allocator.ts`, the top-level `allocator` resolver and its typeDef, and their tests (`allocator.mock.spec.ts`, `hook.spec.ts`, `extract/snapshot.spec.ts`, resolver spec). The `load/index.ts` changes, the legacy Gnosis branch, `load/allocator.spec.ts`, `allocator-security.spec.ts` and the three `docs/allocator-*` files are removed.
>       Provenance: [3609121](https://github.com/yearn/kong/commit/3609121b47798d645ff3426d985cabf03c08c05e), [89d36d9](https://github.com/yearn/kong/commit/89d36d925ab2f87aae77dfa9a8ffc93d47484dd6)

New changes pushed that simplifies the PR significantly per the issue above

murderteeth (changes requested):
The core fix is right. Reading the allocator from the Role Manager at the snapshot block fixes both #471 bugs and removes the forged-event path from #476. A failed read retries the job instead of erasing stored data.

### Issues

- [ ] **packages/web/app/api/gql/typeDefs/allocator.ts:6** — **Breaking change to the GraphQL contract (medium)** — `Allocator.address` changes from `String!` to `String`, so clients typed against the current schema can now get a null address they don't handle. The change isn't needed: the top-level `allocator` query already returns a nullable `Allocator`, so a vault with no allocator can return `null` there.
  Done when: `Allocator.address` is still `String!`, and `allocator(chainId, vault)` for a vault with no assigned allocator returns `null`.
  Provenance: [12c83eb](https://github.com/yearn/kong/commit/12c83eb7)
  Found by: 👤 human

---

## How This Was Reviewed

packages/ingest/abis/yearn/lib/allocator.ts
```diff
@@ -0,0 +1,79 @@
...
+    console.warn('Conflicting allocator ratio interfaces')
+    return emptyRatios()
+  }
+  return pairs[0] ?? emptyRatios()
+}
+
+// The manager comes from the vault read at blockNumber. Required read failures
+// propagate to the snapshot job: no factory fallback and no fabricated clear.
+export async function readVaultAllocator(vault: Address, manager: Address | undefined, strategies: Address[], blockNumber: bigint, rpc: Rpc): Promise<VaultAllocator> {
+  if (!manager) throw new Error('Vault role manager read unavailable')
```
matheus1lva: `!manager` throws and aborts the whole V3 snapshot hook (tvl/apy/fees/composition), not just the allocator fields. `role_manager` is `.optional()` in `SnapshotSchema`, so one flaky sub-call blocks all vault data updates until it recovers. Consider distinguishing absent `role_manager` from a failed read, or keep the prior allocator and continue.

packages/ingest/abis/yearn/lib/allocator.ts
```diff
@@ -0,0 +1,79 @@
...
+      (cause.data !== undefined || cause.signature !== undefined || /^execution reverted\b/i.test(cause.reason ?? ''))
+    if (reverted || cause instanceof ContractFunctionZeroDataError) {
+      console.warn('Allocator assignment unavailable', vault, manager, blockNumber)
+      return { address: null, ratios: {} }
+    }
+    throw error
+  }
+  if (assigned === zeroAddress) return { address: null, ratios: {} }
+  const contracts = strategies.flatMap(strategy => [false, true].flatMap(shared =>
+    ['getStrategyTargetRatio', 'getStrategyMaxRatio'].map(functionName => ({
```
matheus1lva: The `2 interfaces × 2 getters` stride is encoded in three places that must stay in sync: this builder, the getter literal list in the failure filter below, and `[0, 2]` / `index * 4` in `selectAllocatorRatios`. Drift mis-attributes ratios to the wrong strategy silently. Suggest shared `RATIO_GETTERS` / `INTERFACES` consts and derive `CALLS_PER_STRATEGY` + the slice from them.

packages/ingest/abis/yearn/lib/allocator.ts
```diff
@@ -0,0 +1,79 @@
...
+  // allowFailure is enabled. Only getter execution/decoding failures are data.
+  for (const result of results) {
+    if (result.status === 'failure' && (!(result.error instanceof ContractFunctionExecutionError) ||
+      !['getStrategyTargetRatio', 'getStrategyMaxRatio'].includes(result.error.functionName))) {
+      throw result.error
+    }
+  }
+  return { address: assigned, ratios: Object.fromEntries(strategies.map((strategy, index) =>
+    [strategy.toLowerCase(), selectAllocatorRatios(results.slice(index * 4, index * 4 + 4))]
+  )) }
```
matheus1lva: `ratios` is keyed by `strategy.toLowerCase()` but typed as `Record<string, AllocatorRatios>`, so the casing contract is invisible. A caller that forgets `toLowerCase()` gets `undefined` and persists null ratios silently. Suggest exporting `ratiosFor(allocator, strategy)` and using it from the hook.

packages/ingest/abis/yearn/3/vault/snapshot/hook.ts
```diff
@@ -327,13 +306,9 @@ export async function extractDebts(chainId: number, vault: `0x${string}`, strate
...
         : 0n
 
-      const targetDebtRatio = multicall[2]?.result
-        ? Number(multicall[2].result)
-        : undefined
-
-      const maxDebtRatio = multicall[3]?.result
-        ? Number(multicall[3].result)
-        : undefined
+      const ratio = allocator.ratios[strategy.toLowerCase()]
```
matheus1lva: Mixed block semantics in one debt row: these ratios are pinned at `snapshot.blockNumber` (archive client), but the `strategies()`/`performanceFee()` multicall above still runs unpinned at `latest`. Pass `blockNumber` to that multicall (and `rpcs.next(chainId, blockNumber)`) so the row is consistent.

packages/ingest/abis/yearn/3/vault/snapshot/hook.ts
```diff
@@ -97,13 +99,16 @@ type Snapshot = z.infer<typeof SnapshotSchema>
...
   const snapshot = SnapshotSchema.parse(data)
-  const strategies = await projectStrategies(chainId, address, undefined, snapshot)
+  const strategies = await projectStrategies(chainId, address, snapshot.blockNumber, snapshot)
   const roles = await projectRoles(chainId, address)
   if (snapshot.role_manager) appendRoleManagerPseudoRole(roles, snapshot.role_manager)
 
-  const allocator = await projectDebtAllocator(chainId, address)
+  const allocatorData = snapshot.role_manager === zeroAddress
+    ? { address: null, ratios: {} }
+    : await readVaultAllocator(address, snapshot.role_manager, strategies, snapshot.blockNumber, rpcs.next(chainId, snapshot.blockNumber))
```
matheus1lva: The `zeroAddress` ternary duplicates the guard already inside `readVaultAllocator`. Calling the helper unconditionally is equivalent and keeps one source of truth.

packages/ingest/abis/yearn/3/vault/snapshot/hook.ts
```diff
@@ -5,6 +5,7 @@ import { snakeToCamelCols } from 'lib/strings'
 import { EstimatedAprSchema, EvmAddressSchema, ThingSchema, TokenMetaSchema, VaultMetaSchema, zhexstring } from 'lib/types'
 import { parseAbi, toEventSelector, zeroAddress } from 'viem'
 import { z } from 'zod'
+import { readVaultAllocator, type VaultAllocator } from '../../../lib/allocator'
```
matheus1lva: With `projectDebtAllocator` gone, `docs/cpu-cost-analysis.md:25,44,124` and `packages/db/migrations/sqls/20260513233437-cpu-indexes-up.sql:8-10` are stale: both cite it as the reason `evmlog_idx_chain_signature_args_vault` exists. Index now has no reader. Update the notes; consider dropping the index in a follow-up migration.

packages/web/app/api/gql/resolvers/allocator.ts
```diff
@@ -1,25 +1,17 @@
...
-      evmlog
-    WHERE
-      chain_id = $1
-      AND event_name = 'NewDebtAllocator'
-      AND args->>'vault' = $2`,
-    [chainId, vault])
-
-    return snakeToCamelCols(result.rows)[0]
-
+      SELECT ${mergeSnapshotSql()} AS merged
```
matheus1lva: This selects the full merged snapshot blob (debts, composition, sparklines, ...) to read one key. `mergedFieldSql('allocator')` from `lib/mergeSnapshot.ts` projects just that field. `allocator.spec.ts` asserts on `mergeSnapshotSql()` so that assertion needs updating too.

matheus1lva (commented):
Inline comments from an ocr pass (alibaba/open-code-review), triaged. Nothing blocking, mostly medium.

Issue #402, opened 2026-05-09
Switch evmlog source from RPC getLogs to envio Postgres


matheus1lva:
> **Envio-side prerequisites tracked in chain-events/yearn-indexing-test#24** — `transactionIndex` column, chains `100`/`146`/`80094` enabled, per-event coverage audit. Must land before `EVMLOG_SOURCE=envio` rollout.

## Description

Kong currently fetches event logs by calling viem `getLogs()` against configured RPCs (`packages/ingest/extract/evmlogs.ts:41`). Direct RPC calls have been observed to drop logs, leaving holes in the `evmlog` table. The yearn internal envio indexer already maintains a complete, deduped event stream in Postgres. This spec replaces the RPC fetch with a direct pull from the envio DB, gated by an env var so RPC remains a fallback, and adds a maintenance script to backfill historical gaps from envio into kong.

## Context

**Current implementation**
- `packages/ingest/extract/evmlogs.ts:37-48` — `EvmLogsExtractor.extract()` calls `rpcs.next(chainId, from).getLogs({ address, events, fromBlock, toBlock })` to fetch logs.
- Result feeds hooks (`packages/ingest/extract/evmlogs.ts:50-76`) and is queued to `mq.job.load.evmlog` (`:90-95`).
- `packages/ingest/load/index.ts:55-87` upserts each row into `evmlog` and updates `evmlog_strides`.
- `evmlog` schema: `packages/db/migrations/sqls/20240214020032-eventsource-up.sql:1-15`. PK `(chain_id, address, signature, block_number, log_index, transaction_hash)`. `transaction_index int4 NOT NULL`. `topics text[] NOT NULL`. `block_time timestamptz`.

**Target system**
- Envio replica DB at `postgresql://envio_ix_7f3a:…@dpg-d75faobuibrs73bmodug-b.replica-cyan.virginia-postgres.render.com/envio_indexer_jvm1`.
- Schema `envio` holds typed per-event tables (one table per event, e.g. `envio.\"Transfer\"`, `envio.\"StrategyReported\"`, `envio.\"V3RegistryNewEndorsedVault\"`, …).
- Common columns: `chainId`, `vaultAddress` (or registry/factory address), `blockNumber`, `blockTimestamp` (unix int), `blockHash`, `transactionHash`, `transactionFrom`, `logIndex`, plus event-specific args.
- **Missing vs kong** evmlog requirements:
  - No `transactionIndex` column (any event table inspected so far)
[...]

Issue #396, opened 2026-04-24
Vault Allocation History REST API Spec


rossgalloway:
**Date:** 2026-04-24

**Goal:** Add a Kong-owned, vault-scoped allocation history endpoint that serves debt reallocations, strategy allocation states, and DOA annotations from Kong REST cache with minimal frontend processing.

**Architecture:** Kong uses its indexed V3 vault/debt allocator events as the canonical event source, enriches those events with transaction metadata, materializes historical **block-end** allocation states using archive RPC at each relevant block, joins DOA optimizer records as annotations, then serves a cached timeline through a new REST endpoint.

**Tech Stack:** Kong monorepo, Bun/TypeScript, Next.js route handlers, Postgres, Redis/Keyv, viem archive RPC clients, Yearn V3 vault/debt allocator ABIs.

---

## 1. Background

The yearn.fi `add-flow-chart` branch has a product/data spec at:

- `yearn.fi/docs/manual-allocation-reallocation-data-spec.md`

That spec establishes the desired product behavior:

- Show a vault-scoped allocation timeline.
- Explain both DOA optimizer-driven reallocations and manual/operator reallocations.
- Treat on-chain events as canonical.
- Treat DOA records as annotations/proposals, not proof of execution.

Ross confirmed two decisions for the Kong version:

1. Kong has reliable archive RPC access, so Kong should materialize historical block-end states directly.
2. Add transaction metadata needed for classification: `transactionFrom`, `transactionTo`, and `inputSelector`.

## 2. Current Kong Context

Relevant files verified in Kong:

- REST docs: `docs/rest.md`
- REST routes:
  - `packages/web/app/api/rest/snapshot/[chainId]/[address]/route.ts`
  - `packages/web/app/api/rest/reports/[chainId]/[address]/route.ts`
- REST refresh scripts:
  - `packages/web/app/api/rest/snapshot/refresh-snapshot.ts`
  - `packages/web/app/api/rest/reports/refresh.ts`
  - `packages/web/app/api/rest/reports/refresh-historical.ts`
- REST cache helper: `packages/web/app/api/rest/cache.ts`
- DB connection: `packages/web/app/api/db/index.ts`
- Eventsource migration: `packages/db/migrations/sqls/20240214020032-eventsource-up.sql`
- V3 vault snapshot hook and useful projection helpers: `packages/ingest/abis/yearn/3/vault/snapshot/hook.ts`
[...]

PR #481, opened 2026-09-22
Specify vault allocation history and REST delivery

rossgalloway:
### Summary

Kong needs an agreed design for showing how vault allocations change over time. Add a finished-product README, a staged implementation spec centered on the refresh job and REST responses, and real yvUSDC-1 samples for reviewing the data shape.

Related to #396. Event acquisition remains separate work under #402; current allocator assignment corrections are covered by #472.

### How to review

1. Start with the [README](https://github.com/yearn/kong/blob/rg/allocation-history-spec/docs/allocation-history/README.md) to review the intended API behavior and the data frontends use to render charts.
2. Read the [high-level spec](https://github.com/yearn/kong/blob/rg/allocation-history-spec/docs/allocation-history/high-level-spec.md) for the input contract, refresh pipeline, publication rules, and six delivery stages.
3. Use the [sample guide](https://github.com/yearn/kong/blob/rg/allocation-history-spec/docs/allocation-history/samples/yvusdc-1/README.md) to inspect two chart pages and five representative action details. It explains differences between the prototype responses and the proposed contract. The large JSON files are captured responses; begin with the guide rather than reading every line.

### Test plan

- [x] Automated: `git diff --check origin/main...HEAD` passes.
- [x] Automated: all JSON files parse, local documentation links resolve, and all seven captured response checksums match the manifest.
- [x] Capture validation: the recorded September 15 checks cover pagination, chart/detail agreement, and 56 balanced intervals. See `samples/yvusdc-1/validation.json` for the checks and their limits.
- [ ] Manual review: confirm the proposed response and delivery stages meet Kong and Powerglove needs; use the fixed samples for UI integration feedback.

### Risk / impact

Documentation and fixtures only. This PR changes no runtime behavior, database schema, or deployed service. Samples are provisional data generated September 10, 2026; their consistency checks do not certify complete event coverage or independently verify execution classifications. Reverting the documentation commit removes the additions.


murderteeth (changes requested):
## readme questions
- "GET /api/rest/views/allocation-history/:chainId/:address?projection=chart" -- why do we need "projections"
- "limit Number of rebalancing entries; defaults to 25, maximum 100... cursor Continue from pagination.nextCursor, preserving the other query parameters." -- why do we need paging
- "desc for newest first, or asc for oldest first. Defaults to desc." -- why do we need sorting
- refresh
-- "Select a safe processing range" -- what does this mean, why do we need it
-- ideally refresh just catches up when you run it. kiss
-- how long does it take to "catch up" from zero. for one vault, for all vaults
- "...when events or policy records arrive late or are corrected" -- why do we need this. has it been observed happening
- "Each vault has one refresh in progress at a time." -- why is this a requirement
- "Job status records progress, the processed range, and failures independently of the published response." -- why do we need job status records with progress?
- "Pagination and detail links remain tied to their original run while newer runs are published. Kong retains older runs for the supported cursor lifetime; expired references require restarting pagination rather than silently switching to newer data." -- WHAT?? explain
- "Event acquisition and the migration to Envio are separate work, tracked in Kong #402." -- remove from requirements
- Input, "Vault debt, strategy lifecycle, configuration, deposit, withdrawal, and report events" -- what specifically do we mean by "strategy lifecycle", and "configuration". do we need deposit and withdrawal events specifically?
- Input, "Factory deployment evidence" -- why do we need this

## brief reply to answers
- `I'm not sure if there is a question beyond "why do we need need deposit and withdrawal events"`
"strategy lifecycle" is named as an input. that's vague, i want to know what specifically that means. also at a glance it's not clear we need deposits and withdrawals. this important bc 1 kong doesn't currently support full deposit\withdrawal\transfer history, 2 it may be a hint why creating these data structures takes so long. re codex's answer -- that is still vague, i'm not convinced

- `pagination.md`
no paging, it adds more code on both sides with no gain. if performance does become an issue, lets optimizing the datastructure first

- `Catching up from zero: Max time for a vault is about 10min`
why 10 minutes?? what is the bottleneck

- `This is referring to the possibility that the DOA record could come in after the actual onchain actions`
slop, remove

- `To keep the README focused on behavior, I can replace that sentence with: A failed or outdated refresh does not replace a newer successful result`
ok

- `Refresh failures are reported through Kong’s existing logging and monitoring`
ok

- `Looking at the factory may be useful, but maybe not.`
if we dont know or we're not sure, remove. if a concrete reason presents itself, requirements can be updated.

